PR #16, opened 2015-09-10
Overhaul for better compatibility with bower and wiredep.

jeremyzahner:
Hi there!

To support automated integration of devicon into a bower/wiredep/gulp-driven workflow i changed the file structure slightly. Like this, when using bower and task runners like gulp or grunt, it will be much simpler through automated loading of the files (see bower.json's main declaration).


konpa:
Oh! Thanks! I'll have a look at your pull request as soon as possible!

jdrucey:
+1 for this! 
Small changes but would greatly improve the workflow when working with Bower. Otherwise i have to change my sites structure or implement some search and replace functionality at build time.

I know this hasn't been active in a while, but is this still possible to implement?

amacado:
Since this has many conflicts with the current master (no wonder, it's almost 5 years old), I will close this pull request for now. I will indeed create a issue for refactoring this pull request because I think the idea behind it is pretty good.

PR #22, opened 2016-02-10
added slack logos

kriskhoury:
Added slack.com logos. I thought that it should be on the list. Tried my best to follow the guidelines.


konpa:
Thanks very much. I'll have a look as soon as possible!

oligibson:
@kriskhoury sorry for the delay with merging this, I've updated the icons and the font set. We will look to release this soon

PR #36, opened 2016-10-19
Add's LESS Support

oligibson:
- Restructured CSS into LESS files
- Created `gulp less` to build css file


oligibson:
This request is very out of date, I will look into a move to LESS or Sass in the future.

PR #46, opened 2017-01-06
Add various mfizz icons

zacacollier:
Added various linux- & programming language-related icons from [font-mfizz](http://fizzed.com/oss/font-mfizz).

All of these icons come in only one `plain` format, so I refrained from adding `.eps` files in the icon directories despite the stipulation in the contributing guidelines. Nevertheless, let me know if you would like me to include `.eps` files in each directory and I will get on it.


zacacollier:
yep I'll get that to ya tomorrow

Issue #11, opened 2014-11-28
Icon Requests [outdated]


konpa:
_Please comment this issue in order to request an icon and I will update this list._

~~Android~~
~~Amazon Web Services~~
API
AppCode
~~Atom~~
Autodesk
Bash
BEM
BizTalk
Blender3d
~~CakePHP~~
CentOS
~~Chrome~~
Clojure
ClojureScript
~~CoffeeScript~~
chef
CMake
Common Lisp
Compass
Composer
~~CouchDB~~
CUDA
~~C#~~
D
Eclipse
~~ElectronJS~~
~~Elm~~
Elixir
~~Ember~~
~~ExpressJS~~
F#
Firebase
Flask
FreeBSD
~~Gimp~~
Gentoo
Google Developers
Google Cloud
Groovy
Hack
~~Handlebars~~
Haskell
inDesign
~~Inkscape~~
~~IntelliJ~~
iOS
~~Ionic Framework~~
JBoss
Jenkins
Jekyll
Jest
~~JetBrains~~
Knockout
Kubernetes
Kotlin
Lua
Magento
MariaDB
Markdown
Matlab
MaterializeCSS
Maven
Maxthon
Mercurial
Microsoft SQL Server
MyBB
NetBeans
Netscape
~~NPM~~
Objective-C
OpenCL
OpenLiteSpeed
OpenBSD
Opera Mini
OS X
Pascal
Perl
Phalcon Framework
Phoenix Framework
~~PhpStorm~~
Polymer
PostCSS
Prestashop
Pug
puppet
PureScript
QUnit
Qt
~~React~~
ReactiveX
~~Redis~~
Redux
ReSharper
R language
Rust
Scala
Scheme
Scribus
Semantic-UI
SilverStripe
~~Sketch~~
SMACSS
Sphinx
Spring
StackOverflow
Steroids Framework
~~Stylus~~
Sublime Text
SocketIO
SQLServer
Threejs
~~Trello~~
TYPO3
Unity3d
Vaadin
~~Vagrant~~
VB.NET
VirtualBox
~~Vim~~
VimL
~~Visual Studio~~
~~WebStorm~~
Windows Azure
Windows Phone
~~Windows 8~~
Windows 7
Xamarin
~~Yarn~~
Yesod

toastal:
I'll add too some popular things:

#### Server OSs
- CentOS
- FreeBSD
- OpenBSD

#### Servers
- Nginx
- Apache

#### Databases
- MariaDB
[...]

PR #136, opened 2018-08-18
Newicons (c3, embeddedc, labview)

mjlabe:
Added asm, c3, embeddedc, and labview, but I broke the wordmark classes for the icons that don't actually have wordmark icons (like c-plain-wordmark). Could someone please fix that before merging? Why does that exist anyway?


mjlabe:
Will do!

Berkmann18:
@mjlabe Also have a look at #11 where @konpa compiled a huge list of icons and where people commented suggestions.

iamclaytonray:
@Berkmann18 - this is definitely a great list and we can use this as a reference point moving forward but I'd like to keep PRs and issues restricted to a single icon at a time. Trying to manage a list of over 100 icons can be a little challenging and will inevitably become out of sync. With isolated issues and PRs, we can keep things tidy and improve changelogs. Thanks so much!

mjlabe:
Should just add the icons in the PRs and leave the devicons.css alone?

mjlabe:
Finally created PRs #282, #283, and #284. Please review and let me know if anything needs to be changed. Thank you.

amacado:
I will close this pull request and review your new requests! :) Thanks in advance for your contribution @mjlabe

PR #157, opened 2019-02-17
Add clojure icons

razum2um:
clojure-pr


iamclaytonray:
Thank you for adding these icons!

PR #232, opened 2020-06-15
Fix missing d3js glpyh

furby-tm:
It seems like there was a missing glyph for `devicon-d3js-plain`, it was using the glyph for `devicon-d3js-original`. I also corrected the CSS `speak` attribute's CSS2.1 **None** value to be the CSS3 **Never** attribute, however it is deprecated now -- and should be removed.


amacado:
Hello @griffin-copperfield and thank you for your contribution. 
I'm not able to reproduce your reporting about the missing `devicon-d3js-plain`.

Using the instructions like below seems to work as expected.

```
<!-- in your header -->
<link rel="stylesheet" href="https://cdn.rawgit.com/konpa/devicon/df6431e323547add1b4cf45992913f15286456d3/devicon.min.css">

<!-- in your body -->
<i class="devicon-d3js-plain"></i>
```
  
You're right when pointing out there is no `devicon-d3js-original`. We may could add an alias for it because it's pretty the same as the plain version, but thats another topic.

Your hint about the `speak` property is nice, can you point out the documentation/spec where it's marked as depreceated?

Thomas-Boi:
@amacado do you think this PR should be closed? It seems like there are no significant changes and the issue stated here seems to have been fixed.

PR #270, opened 2020-08-26
new icon: Codepen (original, plain, wordmark)

quantik-git:
Added codepen-original-wordmark and codepen-plain( I named it this way since it is only black not sure if it fits exactly the naming convention).


amacado (approved):
We will add an alias to [devicon-alias.css](https://github.com/devicons/devicon/blob/master/devicon-alias.css) once we converted it to a font. Thanks for your work @quantik-git

PR #273, opened 2020-08-29
New Icon: FreeCodecamp (original, original-wordmark)

quantik-git:


amacado (changes requested):
Hello @quantik-git and thanks for your contribution. 

- I'm missing the `plain` and `plain-wordmark` version as svg's. At least the `plain` version is required as mentioned in your contribution guidelines. Since you mention both versions in the `devicon.json` I would expect both to be added to the pull request. 
- The icons do not match our naming convention. Please use `freecodecamp-` instead of `fcc-` because this is required by our build action

I'm also very unsure if FreeCodecamp will match in our set of icons. As pointed out in our project description
> Devicon aims to gather all logos representing development languages and tools

And I'm not sure if FreeCodecamp will fulfill this description. Maybe you can explain why you think it would be a good addition to our set?

quantik-git:
I thought it was a Icon set dedicated to programming related icons not only tools and languages

PR #279, opened 2020-09-02
Created script that will upgrade devicon.json so that objects have the attribute "color" and "aliases"

Thomas-Boi:
Hey guys,

So this branch contains the scripts needed to upgrade the `devicon.json`.

**To Run**

Open a terminal and go to the root directory of the project. Enter `node ./upgradeDeviconJson.js` and the script will be run.

**Note**
* This branch has been rebased over the most recent `master` branch.
* Currently, the script will write its output to a file called `newDevicon.json` in the root directory. This is done so that the original copy doesn't get modified.
* There will be log message about 4 fonts not having colors: `clojure, clojurescript, yunohost, and redux`. The script will then show you their object's attribute. As you can see, they don't have any `font` version for them. Since there are no font, there's no corresponding classes in the `devicon.css` and no colored classes in the `devicon-colors.css`. 
  * We don't have to worry about this for now. To fix this issue, all we have to do is enter some font versions in the `font` attribute of each font then run the build script. 
* There's a commit where I renamed anything `dot-net` related to `dotnet`. This was done so that the folder has the same naming convention as everything else. 

Feel free to checkout the script, it's in `upgradeDeviconJson.js`. I tried to have as much docs as possible but if you have any questions, feel free to reach out.

As always, feedback and suggestions are welcomed.


devicon.css
```diff
@@ -232,7 +232,7 @@
 .devicon-twitter-original:before {
   content: "\e91f";
 }
-.devicon-mocha:before {
+.devicon-mocha-plain:before {
```
amacado, line 235: Can you provide a explanation why you suggest this change?

amacado (changes requested):
> There's a commit where I renamed anything dot-net related to dotnet. This was done so that the folder has the same naming convention as everything else.

I would approve renaming dot-net to follow our conventions, but this would break compatibility. If we want to rename it, I would suggest to provide an alias to make the old version still work.

devicon.css
Thomas-Boi: So I did this because all css classes follow the same pattern: devicon-{techname}-{version}. I found this when I run the upgrade script because mine was failing at mocha.

Thomas-Boi:
Hey amacado,

Please don't merge this branch yet. I'm planning to revert the changes done to dotnet and mocha. I've found a few other fonts that have issues (clojure, redux...etc). I'm thinking that it's better to fix these errors in a later commit/merge. This commit should only focus on the new devicon.json. I think I might also work on the css gulp task and put it in this branch as well.

amacado:
> Hey amacado,
> 
> Please don't merge this branch yet. I'm planning to revert the changes done to dotnet and mocha. I've found a few other fonts that have issues (clojure, redux...etc). I'm thinking that it's better to fix these errors in a later commit/merge. This commit should only focus on the new devicon.json. I think I might also work on the css gulp task and put it in this branch as well.

Good idea to keep things seperated. There is no need to work on the `gulp` task, I already did it in https://github.com/devicons/devicon/pull/281/commits/e6a5f9c567268b3e3b3a53a9877a2bcedc0dd9f6 which you merged in #281. Or do you have something else in mind?

amacado:
@Thomas-Boi Can we delete the branch `upgrade-devicon` and you work on it in your own fork? The less branches we have in this repository the better it is to maintain.

Thomas-Boi:
> There is no need to work on the gulp task, I already did it in e6a5f9c which you merged in #281. Or do you have something else in mind?

The `gulp task` that we have currently relies on the css files staying in the repo. I think that we want to create a system where the css files are generated in memory by gulp using the `devicon.json`. It'd just be an extra step for the `gulp default`.

> Can we delete the branch upgrade-devicon and you work on it in your own fork? The less branches we have in this repository the better it is to maintain.

Sure, I'll delete the `build-integrate` branch as well

amacado:
> The `gulp task` that we have currently relies on the css files staying in the repo. I think that we want to create a system where the css files are generated in memory by gulp using the `devicon.json`. It'd just be an extra step for the `gulp default`.

True, I just wanted to say that we are able to publish this version as a first version of a working build automatisation (including the gulp task generating the *.min.css). Even when our "final" goal of having a single `devicon.json` is not achieved yet (next version of our build automatisation). :)

PR #280, opened 2020-09-04
Action fix, testing, not a real commit

Thomas-Boi:


amacado:
@Thomas-Boi can you delete the `action-fix` branch? You created a lot of branches and I don't wanna end in a mess ;-)

PR #320, opened 2020-10-12
Fix the color css alias and change the repo ref to trigger the workflow

Thomas-Boi:


amacado (approved):

Thomas-Boi:
@amacado,

Just fyi, I went ahead and implemented solution 2 from this [issue](https://github.com/devicons/devicon/issues/321). It was pretty quick and the majority of my time was spent on getting a comment bot to work (which didn't).

The label for this quick test is called "bot:peek". We also need to discuss how to trigger the "bot:build" script if we are going with this method.

PR #328, opened 2020-10-18
fix svg rendering in preview window for showing the svg code 

### LINKED ISSUE #63: Some svg icons don't have colors
Some icons have attribute `fill` (angular's icon for example), but some of them don't. For example, webpack's icon has attribute class `cls-1` instead of `fill`. Is it intended?

### LINKED ISSUE #162: Ruby and Rails SVGs not rendering properly
The Ruby and Rails SVGs do not render properly with the code copied from the box.  However, if I right click on the image and open it in a new tab, it displays perfectly.  If I copy the code from the inspector on that page, it will work fine.  Not only that, if I use the code from the inspector, and the code copied from the box on the frontend, both will display properly.  I think the there is a missing URL somewhere, but I am not very familiar with SVGs.
Ruby SVG is incomplete

### LINKED ISSUE #326: Python SVG Icons Not Working
Hello, so the SVG versions of the Python icon are not working. I tested all of them and they aren't showing in my browser.

amacado:
(ruby, rails, facebook, react, twitter, python,..)

fixes #326; fixes #162; fixes #63

### DIFF
--- a/assets/js/script.js
+++ b/assets/js/script.js
@@ -102,14 +102,9 @@ devicon.controller('IconListCtrl', function($scope, $http, $compile) {
     $http.get(baseUrl + 'devicon.git/icons/' + $scope.selectedIcon.name + '/' + $scope.selectedIcon.name + '-' + svgVersion + '.svg').success(function(data){
 
       var svg = angular.element(data);
-      var paths = svg.find("path");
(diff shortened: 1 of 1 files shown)


Thomas-Boi (approved):
This looks good to me

Issue #334, opened 2020-10-23
Icon request: Haxe


kLabz:
### About the icon

Haxe is a programming language, and having its icon in devicons would allow Haxe icon to be used alongside other programming languages icons in developer tools using devicon. 

### Links
* https://haxe.org/
* https://github.com/HaxeFoundation/haxe
* https://haxe.org/foundation/branding.html

### Preview
![](https://haxe.org/img/branding/haxe-logo-glyph.png)

PR #339, opened 2020-10-25
New icon: Haxe (plain, original)

ezSaldana:
## Added Haxe icon
I've added the plain and original Haxe icons. Give it a check and let me know if there's something I should change.

### Image
Screen Shot 2020-10-25 at 14 58 03

### Issue Reference
#334


devicon.json
```diff
@@ -1242,6 +1242,23 @@
...
+        "versions": {
+            "svg": [
+                "original",
+                "plain"
+            ],
+            "font": [
+                "plain"
+            ]
+        },
+        "color": "#F28820",
```
amacado, line 1259: https://haxe.org/foundation/branding.html

image

```suggestion
        "color": "#EA8220",
```

amacado (changes requested):
Where did you got the color from? I would suggest to go with the official branding color.

devicon.json
ezSaldana: Ok, let me change it.

amacado (approved):

Issue #330, opened 2020-10-19
Icon request: Woocommerce


Clareapumami:
### About the icon
I want to list my skills on my portfolio website using icons. I came across devicon because I was looking for a complete icon set that included Woocommerce. Funny enough, I saw everything else I needed except Woocommerce. It would be nice if you included it in the future.

PR #340, opened 2020-10-25
New icon: Woocommerce (original, original-wordmark, plain, plain-wordmark)

ezSaldana:
## Added Wocommerce icon
I've added the plain, original and wordmarks icons. Give it a check and let me know if there's something I should change.

### Image
Screen Shot 2020-10-25 at 16 49 09

### Issue Reference
#330


amacado (changes requested):
As you can see the icons don't look right when converted to a font. This is probably because you have used multiple colors in it:
`st0{fill:#7f54b3;}.st1{fill:#FFFFFF;}`

You can try this by yourself when uploading the svg to icomoon.


image

ezSaldana:
Ready, give it a check and let me know if it's ok 👍🏽

amacado (approved):
Looking good! Nice addition! 👍🏻 
image

Issue #79, opened 2017-10-11
[Suggestion] Additional brands/technologies


DoronBrayer:
First of all, I love DEVICON. Good job!

Please add the following brands/technologies:

- Stack Overflow
- Go
- iOS
- W3C (World Wide Web Consortium)
- Sublime Text
- Brackets
- CodePen
- JSFiddle
- freeCodeCamp
- Stylus
- PostCSS
- npm (Node Package Manager)
- Yarn
- Ember
- Aurelia
- Redux
- Broccoli
- Mithril.js
- Chart.js
- three.js
- D3.js
- OpenGL
- Direct3D
- PhantomJS
- CasperJS
- Nightwatch.js
- GraphQL
- Riot
- Knockout
- Babylon.js
- KeystoneJS
- Hexo
- Redis Labs
- MongoDB
- Cloud Foundry
- Microsoft Edge
- Microsoft Azure
- Google Cloud Platform
- Electron
- Ionic
- PhoneGap
- Apache Cordova
- Firebase
- Magento
- Flask
- Laravel
- Webpack
- Ghost
- Unreal Engine
- CryEngine
- WebGL
- Unity
- MaxCDN
- KeyCDN
- Cloudflare
- Normalize.css
- Polymer
- Semantic UI
- Material UI
- MDL (Material Design Lite)
- TACHYONS - CSS Toolkit
- Animate.css
- Bulma
- Font Awesome
- Sketch
- Materialize
- Unified Functional Testing [formerly known as HP QuickTest Professional (QTP)]
- Selenium
- ElasticSearch
- Scrapy
- Tornado Web Server
- sqlMap
- Jenkins
- Ansible
- TensorFlow
- certbot

Reference:
https://github.com/kamranahmedse/developer-roadmap
https://github.com/topics
https://github.com/topics/css?o=desc&s=stars
https://github.com/topics/javascript?o=desc&s=stars
https://github.com/topics/html?o=desc&s=stars
https://github.com/topics/python?o=desc&s=stars
https://stackoverflow.com/tags

watchwood:
That's quite a list.  Personally, I'd just be happy to have Craft and Twig in there,

KevinMalo:
+1

Fl4v:
I'd love to see Microsoft Azure being added.

Thomas-Boi:
I know this is an old issue but due diligence: please split the request into individual issue. I'll be closing this for now.

PR #352, opened 2020-10-30
New icon: Materialui (original, plain)

ezSaldana:
## Added Material Ui icon
I've added the plain and original icons. Give it a check and let me know if there's something I should change.

### Image
Screen Shot 2020-10-29 at 18 30 07

### Issue Reference
#79


amacado (approved):
Approved! :) Thanks @ezSaldana 

image

PR #372, opened 2020-12-03
fix build process

amacado:
Currently our font build process fail since it's not ready for `workflow_dispatch` trigger:

image

Since we are only triggering this action manually when a new release should be made we need to specify a fixed name for the resulting branch and can set the base to our master branch.


Thomas-Boi:
That's a weird bug.  Unfortunately, I'm a bit busy until the next week but after that I'll have some free time. I'll get on it ASAP by next week 👍.

PR #378, opened 2020-12-11
Fix a couple issues found in the build script

Thomas-Boi:


Thomas-Boi:
Seems like the issue has been fixed by this commit. I'll post my analysis in the issue linked to this PR.

.github/workflows/build_icons.yml
```diff
@@ -19,10 +19,7 @@ jobs:
           pip install -r ./.github/scripts/requirements.txt
           npm install
       - name: Run icomoon_upload.py
```
amacado, line 21: Since we are not having the `icomoon_upload.py` script anymore, this step should be renamend imo:
```suggestion
      - name: Executing build and create fonts via icomoon
```

amacado (changes requested):
Update looking good to me (not tested yet) only a small suggestion in changing some description 🚀

PR #392, opened 2020-12-13
Moved the index.html of the gh-pages branch into master

Thomas-Boi:
For this branch, I moved all the needed files from the gh-pages branch into the `master`/`main` branch.

This branch `TB_ghPage` is created from the the most recent `develop` branch. This means that whenever we make a new release, we should be able to merge `develop` into `master` with no issue. 

I have tested the site locally and everything seems to work fine. We will need to test what it looks like after it's merged into master though.


amacado:
What about moving it to `/docs` folder (and publishing it as github page). In my opinion it's more structured and looks cleaner than putting all in the root.

Shouldn't this branch be merged with 'develop' first? And from there to the `master` branch?

Thomas-Boi:
I think moving it into the `/docs` folder is a good idea. I will do if for my next commit then.

Regarding merging straight into `master`, I decided to do that because these files aren't needed to be in `develop`. I guess I was trying to keep everything minimal between the branches. Also, the `develop` branch already has an `index.html` and I didn't want the file to be replaced. 

I think there's no issue with merging into `develop` first before `mater`. The addition of a few additional folders should not be too bad. What do you think?

amacado:
I think it's no pain to have some more folders in `development` but it potenially allows us to have a separated "testsite" for `devicon.dev` in the future (based on the dev branch)

Thomas-Boi:
That sounds good. I'll close this PR and open a new one into `develop`.

PR #394, opened 2020-12-13
Moved GitHub pages stuff into a docs/ folder

Thomas-Boi:


amacado (commented):
Is `browserconfig.xml` required? (And in the right place?)

Thomas-Boi:
I actually missed that file, thanks for pointing it out to me 😄 

I did some research on what it does, turns out it's a [Windows thing only](https://danaleegibson.com/what-is-the-browserconfig-dot-xml/)

It doesn't do much to the site but it helps [prevent errors in IE and Windows 8](https://stackoverflow.com/questions/23043429/what-is-a-simple-minimal-browserconfig-xml-for-a-web-site)

So I guess it's not crucial to the site and we can remove it if we want to. As for its location, we just have to put it in the root of the publishing source, so I think in the docs/ folder for our case

Thomas-Boi:
I've just moved the `browserconfig.xml` and `CNAME` file into the `docs/` folder. I think it should be correct since `docs/` is our publishing source. If needed, we can fix it later.

amacado (approved):

Issue #400, opened 2020-12-27
Change icon CSS classes to fill attribute


rawdanowiczdev:
Hello all,
I've been using SVG icons in my portfolio and I encountered a problem with inconsistency of color filling. Most of the icons are properly styled with `fill` attribute but some of them use CSS classes `.cls-1`, `.cls-2` and so on. It creates a scope problem and each next icon using CSS classes overdrives color of the previous ones.

**A bad example (npm icon)**

```html
      <svg viewBox="0 0 128 128">
        <defs>
          <style>
            .cls-1 {
              fill: #cb3837;
            }
          </style>
        </defs>
        <title>Artboard 5</title>
        <path
          id="original-wordmark"
          class="cls-1"
          d="M2,38.5H126V82.21H64V89.5H36.44V82.21H2ZM8.89,74.93H22.67V53.07h6.89V74.93h6.89V45.79H8.89ZM43.33,45.79V82.21H57.11V74.93H70.89V45.79Zm13.78,7.29H64V67.64H57.11Zm20.67-7.29V74.93H91.56V53.07h6.89V74.93h6.89V53.07h6.89V74.93h6.89V45.79Z"
        ></path>
      </svg>
```

**After improvement (npm icon)**

```html
      <svg role="img" viewBox="0 0 128 128">
        <path
          fill="#cb3837"
          d="M2,38.5H126V82.21H64V89.5H36.44V82.21H2ZM8.89,74.93H22.67V53.07h6.89V74.93h6.89V45.79H8.89ZM43.33,45.79V82.21H57.11V74.93H70.89V45.79Zm13.78,7.29H64V67.64H57.11Zm20.67-7.29V74.93H91.56V53.07h6.89V74.93h6.89V53.07h6.89V74.93h6.89V45.79Z"
        ></path>
      </svg>
```

**My solution**
- check all SVG icons one by one
- remove all of CSS Classes
- use `fill` attribute for color
- remove `title` attribute or change to matching one (optional)

I would be happy to fix that issue if there's a need for that.

amacado:
Hello @rawdanowiczdev! You absolutely have a point here. Those "inline css classes" are not very pretty (and have some other problems too - which we have fixed in some workarounds). If you have the time feel free to improve them and create a pull request. I would highly appreciate it!

rawdanowiczdev:
Thanks for your reply! I'll be glad to do so. Currently I have quite a lot of time so you can expect PR before the end of this year.

[...]

PR #403, opened 2020-12-29
improvement: replace svg classes with inline css

rawdanowiczdev:
Hello,
according to [my issue](https://github.com/devicons/devicon/issues/400) I've closed the scope of all icons that were using classes.

With this PR there should be no more conflicts beetween icons in case of using more of them in a single project. Each icon is styled directly with `fill` attribute and if required (for example with gradients) a unique `id`.

I hope it's gonna be helpful.


Thomas-Boi (commented):
The files still retain their original color and shape from the changes. I have only check a small percentage but overall everything seems fine.

Thomas-Boi:
Hey @rawdanowiczdev,

Thank you very much for your work! That seems to be a lot of files to get through so props to your dedication. We will double check your work and accept your PR after more checking.

Out of curiosity, how did you convert the classes to fills? I'm thinking that we might have to look out for this issue in the future and if you have a script that can fix files with these issues, it'd be great if we can use it. 

@amacado this is an interesting problem. I think we might need to check for this issue in future PRs, either manually or using a script. What do you think?

rawdanowiczdev:
Hey @Thomas-Boi,
thanks for fast reply. I did it manually, even though it might seem like a lot of work in fact it wasn't.

- First in Visual Studio Code I searched for `class` keyword in the **/icons** folder.
- Then I did manually go through every icon that were using a CSS class. 
- And finally marking with **CTRL + D** each `class="cls-1"` (and so on) and just simply replacing with `fill="#color"` according to each class.

I believe it would be a good decision to make it mandatory not to use CSS classes for the future icons to avoid any new conflicts.

amacado:
Thanks for your time @rawdanowiczdev! We will need some time to double check all modifications, and make sure everything looks as expected. I would suggest to setup a small github action which will check that future pull requests will pass our requirements of having inline attributes instead of classes for the svg's (we should also add this requirement to our contribution guidelines). A action which auto-corrects those pull requests is optional in my opinion.

I will create a new issue for this request.

Thomas-Boi:
Hey @amacado ,

I'll double check the svg files for us both so we don't repeat each other's work. I've already went through 26 of them so it shouldn't take too long.

Thomas-Boi (changes requested):
I have checked all of the files and most of them are great. However, there are two files that were changed by the replacement process. They are: 

-electron-original.svg
image

ceylon-original-wordmark.svg
image

Besides these two svgs, everything else looks great. We'll accept your PR once these files are updated 😄

rawdanowiczdev:
@Thomas-Boi that's right, I should have seen that! But it's already good to go with my last commit.

amacado (changes requested):
I went trough all icons and it looks fine, except this one:

image

amacado (approved):
Thanks for fixing haskell. For me I'm fine with all changes, waiting for @Thomas-Boi to approve as well :)

rawdanowiczdev:
Nice to hear that! I hope everything's good to go now.

Thomas-Boi (approved):
Good job @rawdanowiczdev 👍 

Everything looks good now. Props to @amacado for spotting Haskell, I didn't notice the color change. 

I'll approve and merge this commit.

PR #408, opened 2020-12-29
[TEST] new icon: testingicon (original, plain)

Thomas-Boi:
This is a test PR to test the new peek script. DO NOT MERGE THIS PR.

PR #412, opened 2020-12-30
 Make the peek script (bot:peek) take screenshot of individual files

Thomas-Boi:
Hello,

This PR will give the peek workflow the ability to take screenshot of individual icons while in the browser. This will make it easier for us to spot issues as well.

You can see an example here at the very bottom of the PR https://github.com/devicons/devicon/pull/408. Once this PR is accepted, I will close that PR as well.


.github/scripts/build_assets/SeleniumRunner.py
```diff
@@ -191,45 +197,46 @@ def test_for_possible_alert(self, wait_period: float, btn_text: str):
...
 
-        try:
             close_btn = self.driver \
                 .find_element_by_css_selector("div.overlayWindow i.icon-close")
             close_btn.click()
         except Exception as e:
             self.close()
             raise e
 
+    def remove_color_from_icon(self):
```
amacado, line 227: Can you explain why this method is required? Which color is beeing removed? :)

.github/scripts/icomoon_peek.py
```diff
@@ -1,33 +1,63 @@
...
+    Find the icon name from the PR title. 
+    :param icons, a list of the font objects found in the devicon.json.
+    :pr_title, the title of the PR that this workflow was called on.
+    :return a list containing the dictionary with the "name"
+    entry's value matching the name in the pr_title.
+    If none can be found, return an empty list.
+    """
+    try:
+        pattern = re.compile(r"(?<=^new icon: )\w+ (?=\(.+\))", re.I)
+        icon_name = pattern.findall(pr_title)[0].lower().strip()  # should only have one match
```
amacado, line 56: I'm not sure if this method of finding the icon in the title is very stable.. We should give some hints (in the comment of the Bot?) to ensure the title of a pull request is right or maybe we can think of a better solution?

.github/scripts/build_assets/SeleniumRunner.py
Thomas-Boi: Hello there,

So this method is to remove any colors found when we upload the svgs up to Icomoon. Sometime, the contributor would want to convert an original svg to an icon then create an alias. Instead of forcing them to remove the colors beforehand, we can use Icomoon's built-in remove color functionality.

For example, React is an icon that use its original version for the icon:
```
    {
        "name": "react",
        "tags": [
            "framework"
        ],
        "versions": {
            "svg": [
                "original",
                "original-wordmark"
            ],
            "font": [
                "original",
                "original-wordmark"
            ]
        },
```

Here's React original:
React Original

There are some plain icons that also have colors in their svgs as well, such as:

Android

So this would allow us to have a clean icon (so it fits within one font symbol) while allowing the plain svgs to retain their colors.

.github/scripts/icomoon_peek.py
Thomas-Boi: I agree with you regarding using the title. I couldn't think of any better way to find the icon that's being included in this commit.

Maybe you can help me think of a better way to do this?

What I've tried:
- Using git diff to find out which folders were added in this PR => upload only those
  - Many PR have multiple commits and git diff needs two different commit SHAs (the current one and the one before the PR) to compare. I couldn't figure out a way to find the commit SHA before the PR.

I'm looking into:
- Using the GitHub Action pull_request context. However, the docs is a bit complicated so I'm not too sure what I can find that's useful.

Why I want to do this:
- Make the peek process faster. The longest step in the workflow is the selenium uploading to Icomoon step. By limiting the amount of files being uploaded, it makes reviewing the PR faster for everyone (usually by about 30 s which I admit is not a lot)
- Let us know which icon to take up close screenshots on. I don't think we should take screenshots for every single icons every time we run the peek workflow since it'd slow down the workflow.

.github/scripts/build_assets/SeleniumRunner.py
amacado: Cool! 👍🏻

amacado: Maybe you can add this as a comment in the code?

.github/scripts/icomoon_peek.py
amacado: Currently I don't have a better solution ready (though about `git diff` as well).. I think it's okey to stick with the title for now. Maybe letting the Bot add a comment which points this out if it could not be found?

.github/scripts/build_assets/SeleniumRunner.py
Thomas-Boi: Yup, for sure!

.github/scripts/icomoon_peek.py
Thomas-Boi: Yeah, that sounds like a good idea. I'll add it in for now and see if we can find a better way. There's something promising in the docs.

Thomas-Boi:
Hi @amacado ,

I've added your suggestions and error handling to this workflow. You can see an example of it in [the test PR](https://github.com/devicons/devicon/pull/408) at the very bottom.

amacado (approved):

amacado:
@Thomas-Boi feel free to merge it once you are ready. We can have this in the v2.7.1 as well.

PR #170, opened 2019-06-24
Added some new icons (Bash, Firebase, Flask, Google-Cloud, Kubernetes, Objective-C, Pandas, Rocksdb, SQLAlchemy, SQL-Server, uWSGI)

ReallyLiri:
Added icons for the following:

- Bash
- Firebase
- Flask
- Google-Cloud
- Kubernetes
- Objective-C
- Pandas
- Rocksdb
- SQLAlchemy
- SQL-Server
- uWSGI

PR #415, opened 2020-12-30
new icon: Bash (original, plain)

### LINKED ISSUE #303: Icon request: Bash
### About the icon
Bash is a super common tool and it'd be great to have a little icon for it. There are already some very nice SVGs that could be used (linked below).
### Links
SVG: https://bashlogo.com/
### Preview
image

EnisMulic:
I took the bash SVG from #170 and cleaned it up in so it doesn't use inline styles, not sure if I did this correctly.


amacado:
Related issue: https://github.com/devicons/devicon/issues/303

devicon.json
```diff
@@ -227,6 +227,22 @@
...
+        "tags": [
+            "shell",
+            "command"
+        ],
+        "versions": {
+            "svg": [
+                "original",
+                "plain"
+            ],
+            "font": []
```
amacado, line 241: We need a font version of the icon to fulfill our https://github.com/devicons/devicon/blob/master/CONTRIBUTING.md guidelines.

icons/bash/bash-original.svg
```diff
@@ -0,0 +1,30 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Generator: Adobe Illustrator 23.0.3, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
```
amacado, line 2: This comment is not required and should be removed.

amacado (changes requested):

Thomas-Boi:
Hi @EnisMulic  ,

Unfortunately, we are running into some issues checking your SVGs. It seems that your forked repo is from an older version of our master branch. In order for us to check your icons, please pull our latest commit from our develop to your fork.

You might find this link useful to update your repo.

Cheers,
Thomas

devicon.json
```diff
@@ -227,6 +227,24 @@
         "color": "#002A41",
         "aliases": []
     },
+    {
```
Thomas-Boi, line 230: It seems that there are two entries for `bash` in the `devicon.json`. Can you please remove one of them please?

Thomas-Boi (changes requested):

EnisMulic:
Any tips to fix this upload issue, it seems to work fine when I do it manually 
image

Thomas-Boi:
Hello @EnisMulic ,

Sorry for the long wait. It seems that we ran into a very peculiar error that we never see before. As you might know, our repo use [icomoon](https://icomoon.io/) to create our icons. This is a browser and not an API, which means that we have to be precise in choosing which element to click etc...

Normally, when we upload icons, we would remove its colors entirely using Icomoon. This is available here:
image

When we upload your icon, we see this instead:
image

This has never happened before so our script didn't know how to handle it. In fact, I would not be surprised if the Icomoon team just upload a new version of their app which might have cause this to happen.

image

Any how, I had to clone your repo and test it locally to fix this issue. I have also seen your icon and it looks great to me. Thus, if you want, we can accept your PR right now. @amacado what do you think? I will open another PR that contain the fix for this issue.

amacado (approved):

Thomas-Boi:
@EnisMulic congratulation, you are our first PR for 2021. Happy New Year and thanks for your work.

PR #416, opened 2020-12-30
Updated CONTRIBUTING

Thomas-Boi:
Hello,

This should be the last update before we release a new version. I've added the new svg requirements and polish up a few sections.


CONTRIBUTING.md
```diff
@@ -16,62 +17,83 @@ First of all, thanks for taking the time to contribute! This project can only gr
 </ul>
 
 <hr>
-<h2 id="overview">Overview on Submitting Icon</h2>
+<h2 id="terms">Terms</h2>
+<p>Here are some terms that we will use in this repo: </p>
+<ol>
+  <li>"Icon" refers to the set of svgs/icons of a technology/tool. Ex: We might refer to the React svgs and React icons as the React Icon</li>
+  <li>"SVG/<code>.svg</code>" refers to the <code>svg</code> versions of the Icons.</li>
+  <li>"icons" refers to the icon versions of the Icons.</li>
```
amacado, line 25: Confusing sentence, what do you want to express? But I like the idea of **Terms**

CONTRIBUTING.md
```diff
@@ -16,62 +17,76 @@ First of all, thanks for taking the time to contribute! This project can only gr
...
+  <li>The naming convention for the svg file is the following: <code>(Icon name)-(original|plain|line)(-wordmark?).</code></li>
 </ul>
 
 <hr>
 <h2 id='orgGuidelines'>Organizational Guidelines</h2>
 <ul>
-  <li>Each icon has its own folder located in the <code>icons</code> folder</li>
-  <li>Each icon folder contains one <code>.eps</code> file and as many <code>.svg</code> files as versions available</li>
+  <li>Each Icon has its own folder located in the <code>icons</code> folder</li>
+  <li>Each Icon folder must contain one <code>.eps</code> file and as many <code>.svg</code> files as versions available (at the minimum one <code>.svg</code> file)</li>
```
amacado, line 89: I'm struggling with the *.eps requirement, do we need this? Or can we make it optional?

Thomas-Boi: I don't think we need the `.eps` for our repo. It seems to be [something for graphic designers ](https://fileinfo.com/extension/eps) and it's probably not needed for our use case. I don't mind removing it honestly but I wonder if any graphic designers want it

amacado: Then let's make it optional and not a "MUST" requirement

CONTRIBUTING.md
Thomas-Boi: I notice that we use "icon" interchangeably for both the technology/tool and the icons of those tools in our repo. So I wanted to define the "Icon" capitalize refers to the tech/tool while the "icon" lower case just means a font icon. I can try to improve the definition for sure.

Thomas-Boi:
I hope the new term definition is clearer. I also made the .eps requirements optional

CONTRIBUTING.md
```diff
@@ -16,72 +17,93 @@ First of all, thanks for taking the time to contribute! This project can only gr
...
+<<<<<<< HEAD
+<ol>  
+  <li>
+    It is not mandatory to have 6 versions for each icon. An icon can only have one or two versions available. Just keep in mind that the minimum is 1 and the maximum 6 (for now). You must also have at least one version that can be make into an icon.
+=======
+<ul>  
+  <li>
+    You don't need to have 6 versions for each icon. An icon can only have one or two versions available. Just keep in mind that the minimum is 1 and the maximum 6 (for now). You must also have at least one version that can be make into an icon.
+>>>>>>> 09c3b64... Updated CONTRIBUTING
+  </li>
```
amacado, line 64: There is something wrong, looks like a merge error?

amacado (changes requested):
`uses: devicons/public-upload-to-imgur@v2.1` is not released yet, but soon so it's ok. Please have a look at the merge error :)

CONTRIBUTING.md
Thomas-Boi: Huh, that's interesting. I did have to cherry pick a commit from another branch (the update imgur ref) so that must have been the cause. I don't understand how my editor would let that pass by me. I'll fix it

Thomas-Boi:
@amacado 

The merge change has been fixed. I also changed some stuff in the python scripts so it fail instead of returning.

Thomas-Boi:
@amacado I've added the fix for #415 in this PR as well since we have to merge this anyway.

amacado (approved):

PR #413, opened 2020-12-30
new icon: Eleventy (original, plain)

tylensthilaire:
Original and plain icons for [Eleventy](https://www.11ty.dev)

Jpeg logo for reference:
8ZfcJCXi_400x400

PR #421, opened 2020-12-31
new icon: testingicon (original, plain) - TEST - DO NOT MERGE 

Thomas-Boi:


Thomas-Boi:
Hi @amacado ,

This is a strange case of bug. This branch here is created from our `origin/develop` branch. The only difference is an extra commit containing the test files and `devicon.json` entry. As you can see, the action works perfectly. 

I'm confused on why the #413 PR is not working. I'll have to dig into it a bit more to see what's causing the error

amacado:
I'm closing this issue to investigate further.

PR #426, opened 2021-01-02
new icon: sqlalchemy (original, original-wordmark, plain)

EnisMulic:
Taken from #170


Thomas-Boi:
Hello,

Here are the results of the peek process:
new_icons
screenshot_0
screenshot_1
screenshot_2

Everything looks good to me. However, I just want to confirm a few things with the amacado regarding the `style="enable-background:new 0 0 128 128;"` and `x="0px" y="0px"`. When everything works out, I will merge your PR.

devicon.json
```diff
@@ -2525,6 +2525,27 @@
...
+            "svg": [
+                "original",
+                "original-wordmark",
+                "plain"
+            ],
+            "font": [
+                "original",
+                "original-wordmark",
+                "plain"
+            ]
```
amacado, line 2544: I would suggest to have `plain` and `plain-wordmark` as font versions and to move the original over to the `aliases` section (https://github.com/devicons/devicon/blob/master/CONTRIBUTING.md#-updating-the-deviconjson-). This result in the advantage to not have the same icon twice in the font file (since the color is removed when beeing transformed to font, see the `bot:peek` screenshots).

amacado (changes requested):

amacado:
@Thomas-Boi can you check if the `aliases` code is correct? I'm always confused with this.. 😆

devicon.json
```diff
@@ -2535,6 +2535,31 @@
...
+        ],
+        "versions": {
+            "svg": [
+                "original",
+                "original-wordmark",
+                "plain"
+            ],
+            "font": [
+                "plain",
+                "plain-wordmark"
```
Thomas-Boi, line 2552: You don't have to put the "plain-wordmark" here. Since there's no "sqlalchemy-plain-wordmark.svg", this would go into the `aliases` attribute

devicon.json
```diff
@@ -2535,6 +2535,31 @@
...
+            ]
+        },
+        "color": "#333333",
+        "aliases": [
+          {
+            "base": "plain",
+            "alias": "original"
```
Thomas-Boi, line 2559: You don't have to create an `alias` of the "plain" for the "original". We create aliases due to how the versions are used by the community. "plain" versions are guaranteed to be Icons while "original" aren't. Thus, people would expect that there's a class name like `.sqlalchemy-plain` to be available while don't think much of the `.sqlalchemy-original` class.

Since you already have the "plain" version, there is no need to create the "original" version. You can remove the alias object here.

Second thing: if you want to create a "plain-wordmark" version, you should put it here. This is applicable as long as the "original-wordmark" can be made into a font version.

```
"aliases": [
  {
    "base": "original-wordmark",    
    "alias": "plain-wordmark",
  }
]
```

Thomas-Boi (changes requested):
So I started a review about the `aliases` attribute base on what we currently use it for. However, it gets me thinking about whether this attribute is really necessary. I'll open a discussion on this topic to get everyone's feedback. Until then, we can either wait for this to resolve (maybe a week depends on what kind of feedback we get) or we can merge it when you fix the `devicon.json`.

Thomas-Boi:
Hey @EnisMulic thanks for the changes. Sorry for not merging this earlier, it kinda escaped my mind 😅

EnisMulic:
It's okay. I didn't update the PR right away anyway. 
Do you mind closing #170 as this was the last icon from that batch.

Thomas-Boi:
Sounds good, I'll close that PR 👍

PR #440, opened 2021-01-03
fix typo in imgur output

amacado:
I think the build failes because of a typo in the output parameters: https://github.com/devicons/devicon/runs/1638047227
image


Thomas-Boi:
Thank you for fixing that 👍

PR #451, opened 2021-01-03
fix icon: bash (plain)

EnisMulic:


Thomas-Boi:
I think this change should be ok to merge into develop. This is the raw link for the [object](https://raw.githubusercontent.com/EnisMulic/devicon/EnisMulic/fix/icon/bash/icons/bash/bash-plain.svg)

We can't really test it with our website either until we merge it. @amacado what do you think?

amacado:
Looking good to me, lets give it a try

PR #454, opened 2021-01-04
Change the peek workflow to use upload artifacts

Thomas-Boi:
Hi,

This PR has the same code as the PR in #452 .While I changed the `peek_icon.yml`, I still keep the old one in the `.github/build_assets/drafts` in case we ever get it to work.


amacado (approved):
It's better than nothing, thanks! 👍🏻

Issue #444, opened 2021-01-03
SVG extraction failes on icons containing a comment at the top (f.e. bash-plain.svg)


amacado:
EDIT: See https://github.com/devicons/devicon/issues/444#issuecomment-753699913 for explaination

There is a problem with displaying the `plain` version of the new bash icon. It shows up but not displaying the svg code below:
It works fine for the original version and also for the font versions. I wonder if this might be a result of the updated build process (removing color function or so? It's just a guess).

![image](https://user-images.githubusercontent.com/8781699/103469596-1f7a2180-4d67-11eb-92bb-f3812db1185c.png)

Interesting fact: When you open the svg in the browser it will show repeated:
https://raw.githubusercontent.com/devicons/devicon/master/icons/bash/bash-plain.svg
![image](https://user-images.githubusercontent.com/8781699/103469605-3de01d00-4d67-11eb-9ab6-5b1bd2064528.png)

Thomas-Boi:
Hmm that's strange. The Icomoon color removal only applies to the font. The original SVG is not affected in anyway.

amacado:
This was just a guess on my side, it is probably another reason..

EnisMulic:
@amacado I just checked and it looks like there was some extra svg elements in the original file I took, removing that seems to fix the issue
https://raw.githubusercontent.com/EnisMulic/devicon/fix/icon/bash/icons/bash/bash-plain.svg
![image](https://user-images.githubusercontent.com/17247763/103488664-cfef3080-4e0e-11eb-8d63-c36b3b8db205.png)

Thomas-Boi:
@EnisMulic I was wondering about that too. It seems that the plain version somehow has more content than the original. 

Can you create a new PR that contain this fix please?

amacado:
The error is still there. I will investigate further. Think it's related to the way we extract the svg to show the code.

amacado:
Yep, the error is related to the way we extract the information from the svg. When there is a comment block at the beginning of the svg we fail in showing the correct code. Some examples:

![image](https://user-images.githubusercontent.com/8781699/103492519-c32d0580-4e2b-11eb-8313-7a9d0fca0377.png)
![image](https://user-images.githubusercontent.com/8781699/103492540-f1124a00-4e2b-11eb-83ab-4be12c5fdc83.png)
![image](https://user-images.githubusercontent.com/8781699/103492549-fd96a280-4e2b-11eb-8940-d2560b98081d.png)

[...]

PR #455, opened 2021-01-04
fix displaying svg code in devicon.dev

amacado:
Will close https://github.com/devicons/devicon/issues/444#issuecomment-753699913.

While fixing this bug in #444 I noticed some svg do not fulfill our viewbox requirement (128x128) and therefore the extracted code is wrong. I will create a extra issue for this bug.


Thomas-Boi (approved):
Hey amacado,

This PR looks good. I'm glad that you were able to catch it and fix it. I'll approve this PR

PR #66, opened 2017-06-25
added flask icon

moghya:
flask original and wordmark

PR #463, opened 2021-01-05
new icon: flask (original, original-wordmark)

EnisMulic:
Taken from #170


amacado:
@EnisMulic maybe you can have a look at this https://github.com/devicons/devicon/pull/66? Maybe we can use them to enrich this pr?

icons/flask/flask-original-wordmark.svg
```diff
@@ -0,0 +1,201 @@
+<?xml version="1.0" encoding="utf-8"?>
+<svg version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink"
```
amacado, line 2:
```suggestion
<svg version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink"
```

amacado (changes requested):
Removing the `id="Layer"` and it's fine I think. Maybe @Thomas-Boi can have a look at the `aliases`?

new_icons
screenshot_0

screenshot_1

Thomas-Boi:
@amacado I think @EnisMulic is waiting for us to decide whether to keep using the `aliases` attribute in our repo. I've opened a discussion on this topic, which you can see [here](https://github.com/devicons/devicon/discussions/465)

That being said, @EnisMulic got the `aliases` right. Everything looks good and we can just merge this PR. If we need to fix the `aliases`, we'll have to do it for the whole repo later anyway.

Issue #197, opened 2020-03-20
Icon request: default programming related icon when an technology is missing


unixconky:

Thomas-Boi:
Besides looking for a generic icon, we have to figure out how to do this. The best way the I can think of is adding this line to the top of the `devicon.css` file:

```

[class^="devicon-"]::before, 
[class*=" devicon-"]::before {
  content: "\e93a";
}
```

Since CSS cascade down, icons that already exist would override this declaration. If it doesn't, the default content will be displayed (in this case the devicon icon).

I will look more into this issue. If you have any idea on what kind of icon you would like to see as the placeholder, let me know. Otherwise I'm using the Devicon icon.

amacado:
I'm not sure if this is a good idea. I understand the request but it could be realy confusing (I don't know any other icon font which is doing this kind of thing). I'm asking myself: What is the advantage of having a placeholder for a technology? What should it express? I can't see any use-case where this would make sense. Happy to hear other opinions about that.

Thomas-Boi:
When I made the feature, I was thinking of situations like:
-Someone wants to auto-generate icons in their website but the icons might not exist in our repo
-Maybe they make a typo and the site now looks weird without the icon.

That being said, both of these use cases can be easily fixes by testing the site before deployment.

Thomas-Boi:
I'm thinking that we should close this issue and the related PR. I don't think there's any reason for us to keep this feature. It might also slow down the browser since the default icon CSS selector would have to do extra checking. What do you think @amacado ?

amacado:
I agree!

PR #469, opened 2021-01-07
Added a script that will create a default icon fall back for all devicon classes

Thomas-Boi:
This script will add a default icon to all classes that start with `devicon-` . Currently, it is set to the devicon-plain-wordmark icon. We can use any icons, as long as we have it in the repo.

EDITED: You can see a test view in the `.github/test.html`


amacado (changes requested):
I would like to wait for feedback on https://github.com/devicons/devicon/issues/197#issuecomment-756415440 before continuing here.

amacado:
@Thomas-Boi you decide if you wanna keep the branch (in your fork), but please remove it from this repo, ok? :)

Issue #478, opened 2021-01-09
Resource not accessible by integration error


Thomas-Boi:
This error is generated by the comment step in the `Check SVGs` workflow:

![image](https://user-images.githubusercontent.com/43018778/104106590-f4626700-526b-11eb-92dc-bf329eef9886.png)

The interesting thing is this step works fine if the PR is opened by a branch inside the repo:

![image](https://user-images.githubusercontent.com/43018778/104106696-24aa0580-526c-11eb-9baf-a22de44d1665.png)

I will look more into this. This seems like another issue with access or auth.

pan3793:
Encounter same issue on fork repo, do you find the solution?

amacado:
@pan3793 currently @Thomas-Boi is working for finding a solution. This is more like a GitHub actions limitation and not a specific devicon problem.

There is f.e. this comment action https://github.com/mshick/add-pr-comment/ which solves this limitation by creating a proxy service for commenting on forked branches. But this requires to have a proxy up and running, which needs to be maintained.. Maybe we can find a simpler solution for this.

Thomas-Boi:
So I poke around a bit more. It seems that the repo you mentioned was correct. No matter which Action I used, it got blocked whenever it's a fork. After doing some more research, the reason and solution can be found [here](https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/). However, this requires a private repo, which obviously can't be done for us.

I have seen an interesting way to by pass this issue. @nyurik (see [here](https://github.com/nyurik/auto_pr_comments_from_forks) create a cron job/`schedule` workflow that runs every 5 mins and checks whether there's a new PR. If there is, find the workflow associated, download the artifacts, do its thing, then post on the PR. This is actually the inspiration for [my post](https://github.com/devicons/devicon/discussions/383#discussioncomment-270765)

I'll keep digging and see what I can come up with. It seems like many people in the community have this complaint too and GitHub is looking into it. Further updates will probably go into the [discussion](https://github.com/devicons/devicon/discussions/383) since this is bigger than the `check_svg` workflow.

pan3793:
Found a simple solution to solve this, see detail at https://github.com/housepower/ClickHouse-Native-JDBC/pull/294

Thomas-Boi:
Hi @pan3793 ,

I actually saw this resource when I was researching for a solution. At first, I was hesitant about this because of something that I found in the blog:

[...]

PR #479, opened 2021-01-09
Check svg upgrades

Thomas-Boi:
This PR introduces some new features and bug fix to the `check_svg` workflow.

- Refactor it into two workflows. 
  - `check_svg_monthly` will be run monthly to ensure all the svgs in the repo is up to standard. This is to prevent cases where old svgs is out of date without us knowing.
  - `check_svg_on_pr` will only check the svgs in a PR. There's an Action that let us know which svgs were added in a PR which is incorporate in this. If all goes well, we can use this action in the `peek_icon` workflow as well. 
- All files now log an error's traceback.
- Bug Fix: the comment bot of the `check_svg_on_pr` now use a new action. There were issues in the old action (see #478) that I hope this will fix.


Thomas-Boi:
Seems like the bug fix doesn't actually work. The new action has the same issue as the old one. Don't merge this in yet, I'm trying to find an alternative.

PR #491, opened 2021-02-15
new icon: r (original, plain)

jakob-r:
image

Icon for the R programming language: https://en.wikipedia.org/wiki/R_%28programming_language%29


amacado:
Hello @jakob-r for contributing `R`. I know this icon was requested by many people.
I fixed the title of this pull request to make our `bot:peek` command work (a space was missing, so it failed on first try).  @Thomas-Boi: It also [failed a couple of times](https://github.com/devicons/devicon/runs/1909067471) because of the known `No connection could be made because the target machine actively refused it. (os error 10061)` error). I will try it again later this day.

amacado:
I uploaded the icons myself to check them, here are the results:

image

`plain `and `original ` look very similar, but they still differ so I think it's fine to add them both.

Both icons are a little off-centered (top-bottom):
image
@jakob-r maybe you can have a look at that?

amacado (changes requested):
Requesting change to fix the top/bottom spacing.

Thomas-Boi:
Hi @amacado,

I don't know what's causing these errors. It seems so random that I feel it might be a multi-threading or multi-processing issue. I'll search around and see if this has happened before. 

At the very least, we know that the new workflows work 😄.

amacado:
> I don't know what's causing these errors. It seems so random that I feel it might be a multi-threading or multi-processing issue. I'll search around and see if this has happened before.

@Thomas-Boi I just wanted to let you know that this error still occurs sometimes ;-) My guess is that it depends on the server which executes the action. I assume that icomoon blocks an ip (f.e. when many requests occur from the same machine in a short period) and since we are sharing the server where the action is executed (i guess?) this error could be caused by other actions which run independly from our action.. This would be an explaination why this error seems to occur randomly.

Thomas-Boi:
>  I assume that icomoon blocks an ip (f.e. when many requests occur from the same machine in a short period) and since we are sharing the server where the action is executed (i guess?)

Oh right, I forgot that this was our working theory for now. I wonder if we should write another script that retry the connection. Let's just keep an eye on this for now.

jakob-r:
> Requesting change to fix the top/bottom spacing.

So you want it centered or is there a certain baseline? I tried to center the circle to give it a more balanced look.

jakob-r:
> I uploaded the icons myself to check them, here are the results:
> 
> image
> 
> `plain `and `original ` look very similar, but they still differ so I think it's fine to add them both.
> 
> Both icons are a little off-centered (top-bottom):
> image
> @jakob-r maybe you can have a look at that?

On the left my plain version - wher I substracted a part of the circle so the R becomes more visible. I was hoping this kind of modification would be ok?
The right is original just without colors.
So from my point of view this would be ok.

amacado:
@jakob-r feel free to re-request a review once you commited the centering of the icon :)

jakob-r:
> @jakob-r feel free to re-request a review as soon as you commited the centering of the icon :)

I centered the icon as requested.

PR #495, opened 2021-02-20
new icon: fsharp (original)

maltejur:
I added a icon for [F Sharp](https://fsharp.org/). Just the original version, because it is simple enough (I hope I wrote the devicon.json the right way for that, first PR here so I am still learning how it works).

Screenshot 2021-02-20 160910


amacado (changes requested):
Hello @maltejur! Thanks for contributing `F#`. The icon looks great, but to fulfill our [guidelines](https://github.com/devicons/devicon/blob/master/CONTRIBUTING.md):

> The plain and line versions (with or without wordmark) need to stay as simple as possible. They must have only one color and the paths are united. We will strip the color when turning it into icons so they can have any color.

I would like to ask you to add a plain version of the icon to use it as font. You can simple remove the color from the `original` icon and save it as `fsharp-plain.svg`. So we can use it in the font. This ensures each font icon has at least a plain or line version (naming convention).

maltejur:
Sure
> We will strip the color when turning it into icons so they can have any color.

That line confused me a little bit, I thought this was an automated process, but now I also added a plain icon.

amacado (approved):
>That line confused me a little bit, I thought this was an automated process, but now I also added a plain icon.

Yeah, sorry. I understand that this can be confusing.. Currently our bot does not support to fix this by itself and reupload the result back to the repo. We are working on this.. :)

Anyway: `fsharp` icons looking great! Thanks for adding them to our project @maltejur. They will be included in our next release! :)

PR #497, opened 2021-02-21
automate workflow for npm publish

amacado:
I created a workflow which will be triggered when a new release is created. The workflow will try to publish the latest version to [`npm`](https://www.npmjs.com/package/devicon).

When we want to create a new release, we need to update the `package.json` and `package-lock.json` and set it to the new release version. This can be done via the command (replace v2.9.0 with the new version):

```
npm version v2.9.0 -m "bump npm version to %s"
```

This should be done **before** `development` is merged into `master` and a new release is created. Otherwise the workflow will fail and throw this exception:

```
403 403 Forbidden - PUT https://registry.npmjs.org/devicon - You cannot publish over the previously published versions: xxx.
```

When this error occurs you need to checkout the branch, bump the version in the `package.json` and either publish the package manually or creating a new release to trigger the workflow again.


Thomas-Boi (approved):
Hey amacado,

The code looks good. I briefly wonder if we should make an action to automate this but we probably won't do releases that often to warrant that.

Good job on this feature. I totally forgot about it 😅

PR #500, opened 2021-02-21
new icon: elixir (original, original-wordmark, plain, plain-wordmark)

maltejur:
Original, plain and wordmark icons for [elixir](https://elixir-lang.org/). There already is #216, but that seems to be stale.

elixir-original-01
elixir-original-wordmark-01
elixir-plain-01
elixir-plain-wordmark-01


amacado (approved):
Thanks for adding `elilxir` @maltejur! It will be included in our next release! :)

PR #509, opened 2021-02-22
Updated the logging of the actions

Thomas-Boi:
Hi,

This PR intends to fix the logging for the Peek Bot Action:

image

And provide the path name for the Check SVG Action. Right now, it's just the name:
image

image

Now:
image

This should print out the whole path:
image
 

EDIT:
This now contains a fix to the `post_check_svgs_comment.yml` workflow that will run when there are no SVGs to check and post the wrong message (see the below image and GitHub message):
image

The fix:
- Fix the above issue so it don't post the wrong message
- Post no message if there's no SVG to check (it's getting annoying 🙃)

**NOTE**: The change will only take effect when the yaml file is merged into master (recall the workflow_run restriction) so it'd be great if we can put this in the next release.

Cheers,


.github/workflows/post_check_svgs_comment.yml
```diff
@@ -36,13 +37,11 @@ jobs:
 
       - name: Comment on the PR about the result - Success
         uses: jungwinter/comment@v1 # let us comment on a specific PR
-        if: success() && steps.err_message_reader.outputs.content == 'None'
+        if: success() && steps.err_message_reader.outputs.content == '1'
```
amacado, line 40: I think we should use the enum here? 
```suggestion
        if: success() && steps.err_message_reader.outputs.content == SVG_STATUS_CODE.SVG_OK.value
```

.github/workflows/post_check_svgs_comment.yml
```diff
@@ -54,7 +53,7 @@ jobs:
 
       - name: Comment on the PR about the result - SVG Error
         uses: jungwinter/comment@v1 # let us comment on a specific PR
-        if: success() && steps.err_message_reader.outputs.content != 'None'
+        if: success() && (steps.err_message_reader.outputs.content != '0' || steps.err_message_reader.outputs.content != '1')
```
amacado, line 56: I'm not 100% sure why, but this condition seems not to work since the bot is now posting a message even when the content is 0. 

Also I suggest to reuse the enum here:
```suggestion
        if: success() && (steps.err_message_reader.outputs.content != SVG_STATUS_CODE.NO_SVG.value || steps.err_message_reader.outputs.content != SVG_STATUS_CODE.SVG_OK.value)
```

amacado (commented):
I did not tested my code suggestions, those are just my thoughs about it.

.github/workflows/post_check_svgs_comment.yml
Thomas-Boi: I wanted to use the enum but the files are different. The enums are defined in a python file while the workflow is in YAML. I don't think there's a way to import it since they are different.

.github/workflows/post_check_svgs_comment.yml
Thomas-Boi: The reason for content is '0' but the script fails is due to the workflow-run restriction. 

image


The change will only take effect when the yaml file is merged into master. You can test this by creating a new step in the workflow file like this:

```
- name: Test
  run: echo hello world
```

Then check the Actions tab. The new step will not show up even though the syntax is valid.

amacado: Thanks for the explanation!

amacado (approved):
@Thomas-Boi thanks for improving the script! I approve this pull request!

.github/workflows/post_peek_screenshot.yml
```diff
@@ -54,7 +54,7 @@ jobs:
             I'm Devicons' Peek Bot and I just peeked at the icons that you wanted to add using [icomoon.io](https://icomoon.io/app/#/select).
             Here is the result below:
 
-            ![Peeked Icons (top left)]({0})
+            {0}
 
             Here are the zoomed-in screenshots of the added icons:
```
Panquesito7, line 59: When you add more than two icons for the font versions, it adds up an `\n` in the GitHub comment:

image

Panquesito7: Has this been fixed yet? Or is it fixed in this PR? 😄

amacado: @Panquesito7 a few seconds too late before. I will create a new issue based on your report! Thanks! 👍🏻

Issue #510, opened 2021-02-22
When you add more than two icons for the font versions, it adds up an `\n` in the GitHub comment


amacado:
When you add more than two icons for the font versions, it adds up an `\n` in the GitHub comment:

![image](https://user-images.githubusercontent.com/51391473/108760642-87511b00-7513-11eb-8fb4-02e7d735d1ee.png)

_Originally posted by @Panquesito7 in https://github.com/devicons/devicon/pull/509#discussion_r580538363_

Panquesito7:
I think the problem is in the script, isn't it?
https://github.com/devicons/devicon/blob/825033326bb8256e56556fa85346bb7ae9e34d14/.github/scripts/icomoon_peek.py#L32-L33

Thomas-Boi:
> I think the problem is in the script, isn't it?
> https://github.com/devicons/devicon/blob/825033326bb8256e56556fa85346bb7ae9e34d14/.github/scripts/icomoon_peek.py#L32-L33

It's not this one unfortunately, it's this one:

![image](https://user-images.githubusercontent.com/43018778/108800867-22181c80-7549-11eb-909c-1ec965f4cf94.png)


The first one is for logging inside the Python script. The second one is the Action that actually posting the message.

Panquesito7:
Ah, OK. I see. I think the best to do is to remove the `\n`.

amacado:
Fixed in https://github.com/devicons/devicon/pull/511

PR #511, opened 2021-02-23
Remove \n from script

Thomas-Boi:
This should fix the issue #510

### DIFF
--- a/.github/workflows/post_peek_screenshot.yml
+++ b/.github/workflows/post_peek_screenshot.yml
@@ -47,7 +47,7 @@ jobs:
         if: success()
         env:
           OVERVIEW_IMG_MARKDOWN: ${{ fromJSON(steps.icons_overview_img_step.outputs.markdown_urls)[0] }}
-          DETAILED_IMGS_MARKDOWN: ${{ join(fromJSON(steps.icons_detailed_img_step.outputs.markdown_urls), '\n') }}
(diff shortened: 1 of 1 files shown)


.github/workflows/post_peek_screenshot.yml
```diff
@@ -47,7 +47,7 @@ jobs:
         if: success()
         env:
           OVERVIEW_IMG_MARKDOWN: ${{ fromJSON(steps.icons_overview_img_step.outputs.markdown_urls)[0] }}
-          DETAILED_IMGS_MARKDOWN: ${{ join(fromJSON(steps.icons_detailed_img_step.outputs.markdown_urls), '\n') }}
+          DETAILED_IMGS_MARKDOWN: ${{ join(fromJSON(steps.icons_detailed_img_step.outputs.markdown_urls), ' ') }}
```
Panquesito7, line 50: Why not just

```suggestion
          DETAILED_IMGS_MARKDOWN: ${{ join(fromJSON(steps.icons_detailed_img_step.outputs.markdown_urls), '') }}
```
?

Panquesito7 (dismissed):
👍 apart from the above.

.github/workflows/post_peek_screenshot.yml
Thomas-Boi: Oh, I was unsure what will happen if two image markdowns are next together. Now that you pointed it out, I just tested it and it seems like there's no issue.

Panquesito7 (approved):
Looks good. Can be merged. 👍

amacado (approved):

Issue #514, opened 2021-02-23
"Post Check SVG" action is not working as expected


Thomas-Boi:
In PR #513, the "Post Check SVG" action still post the wrong message. This is strange since the `master` branch has been updated with the new version. 

Here is the old version with the incorrect name before we merge into `master`:
![image](https://user-images.githubusercontent.com/43018778/108892372-a22f9800-75c4-11eb-900f-bd2aca892ddb.png)

Here is the new version:
![image](https://user-images.githubusercontent.com/43018778/108892406-ac519680-75c4-11eb-83cd-85aa6b264f2b.png)

This name change was made in #509 and proves that the file has been merged into master. 

I'll look more into this issue during the weekend and test it with my forked repo.

amacado:
There is also this failed action https://github.com/devicons/devicon/actions/runs/593486887 

![image](https://user-images.githubusercontent.com/8781699/108892744-89e56a80-7610-11eb-8041-c64cfbd5e662.png)

Thomas-Boi:
> There is also this failed action https://github.com/devicons/devicon/actions/runs/593486887

This was bad design on my part. I'll have to change it later. Pretty much, what's happening is I only check for 2 results: successful run and failure.

However, there is also a third outcome: the workflow was skipped since the label is not "bot:peek". This is getting lumped into that.

There's also the bad status checks for getting the PR number. I'll have to change it to "always()" so the script will always have a PR to to.

Panquesito7:
This issue is already fixed. Closing.

PR #520, opened 2021-02-24
Fix issue with "post_peek_screenshot" action failing when it's not supposed to

Thomas-Boi:
This PR should fix the issue mentioned in #514 .

What's new:
The "post_peek_screenshot" will now check for three different states of the trigger "peek_icons" action: "skipped", "success", and "failure. These values have to be found by console logging the `github.event` object. The docs don't mention this for some reason.

image

Also, GitHub Actions will construct env variables before an if check by default:
image

Before, there was an env variable that relies on a skipped step => script was failing since it couldn't construct this variable.

That's pretty much it. I've tested it using my forked branch. The behaviours are:
- If "peek_icons" was skipped (label incorrect), the "post_peek_action" would also skip
- If "peek_icons" was successful, the "post_peek_action" would post successful message
- If "peek_icons" failed (either due to user or system error), the "post_peek_action" would post failure message. It'll be up to us to find what's wrong. 

I thought about separating user and system error but I don't think it's worth it.  It'll require some extra code and the benefit (in my opinion) is not worth it. If I was to do it, it'll be something like the check SVG workflow.

Let me know what you think

### DIFF
--- a/.github/workflows/peek_icons.yml
+++ b/.github/workflows/peek_icons.yml
@@ -3,7 +3,10 @@ on:
   pull_request:
       types: [labeled]
 jobs:
-  build:
+  peek:
+    # four outcomes: successful check and upload,
+    # unsuccessful check (fail due to user),
+    # fail due to system, skipped
     name: Peek Icons
     if: github.event.label.name == 'bot:peek'
     runs-on: windows-2019
@@ -20,6 +23,18 @@ jobs:
           python -m pip install --upgrade pip
           pip install -r ./.github/scripts/requirements.txt
 
+      - name: Save the pr num in an artifact
+        shell: bash
+        env:
+          PR_NUM: ${{ github.event.number }}
+        run: echo $PR_NUM > pr_num.txt
+
+      - name: Upload the pr num
+        uses: actions/upload-artifact@v2
+        with:
+          name: pr_num
+          path: ./pr_num.txt
+
       - name: Run icomoon_peek.py
         env:
           PR_TITLE: ${{ github.event.pull_request.title }}
@@ -35,18 +50,6 @@ jobs:
         with:
           name: screenshots
           path: ./screenshots/*.png
-
-      - name: Save the pr num in an artifact
-        shell: bash
-        env:
-          PR_NUM: ${{ github.event.number }}
-        run: echo $PR_NUM > pr_num.txt
-
-      - name: Upload the pr num
-        uses: actions/upload-artifact@v2
-        with:
-          name: pr_num
-          path: ./pr_num.txt
 
       - name: Upload geckodriver.log for debugging purposes
         uses: actions/upload-artifact@v2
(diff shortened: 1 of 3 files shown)


.github/workflows/peek_icons.yml
```diff
@@ -20,6 +23,18 @@ jobs:
           python -m pip install --upgrade pip
           pip install -r ./.github/scripts/requirements.txt
 
+      - name: Save the pr num in an artifact
```
Panquesito7, line 26:
```suggestion
      - name: Save the PR number in an artifact
```

.github/workflows/peek_icons.yml
```diff
@@ -20,6 +23,18 @@ jobs:
...
+        shell: bash
+        env:
+          PR_NUM: ${{ github.event.number }}
+        run: echo $PR_NUM > pr_num.txt
+
+      - name: Upload the pr num
```
Panquesito7, line 32:
```suggestion
      - name: Upload the PR number
```

.github/workflows/peek_icons.yml
```diff
@@ -20,6 +23,18 @@ jobs:
...
+        uses: actions/upload-artifact@v2
```
Panquesito7, line 33: I think it's better to use the `master` version here. It's more updated than the `v2` branch.
```suggestion
        uses: actions/upload-artifact@master
```

Panquesito7:
Weird, not sure why here it displays two `Peek Icons`. 🤷‍♂️

image

At the bottom, it only displays one. 🤔

.github/workflows/peek_icons.yml
Thomas-Boi: I think it might be better to stay with v2 or at least a specific tag. That way, if the upload-artifact repo is updated and breaking change occurs, we don't have to worry about it until we upgrade the action.

Thomas-Boi:
> Weird, not sure why here it displays two `Peek Icons`. 🤷‍♂️
> 
> At the bottom of the PR, it only displays one. 🤔

Oh, I know this one. Currently, our peek action is triggered by the "bot:peek" label. To do this, we create an event whenever a label is added and check the name:

image

If the name doesn't match, it skips. 

Currently, we have 2 labels. Since they aren't "bot:peek", the action will be skipped, thus the above situation.

Panquesito7 (approved):
Looks good. Thank you! 🎉 👍 
Let's wait on @amacado's approval. 🙂

.github/workflows/peek_icons.yml
amacado: @Thomas-Boi is right, (almost) always stick to a specific version and do not target a `master` or `latest` branch. Targeting a `latest`-release is not secure in terms of code injection attacks. You might wanna take a look at https://github.com/devicons/devicon/issues/395.

amacado (approved):
Good explanation @Thomas-Boi!

PR #542, opened 2021-03-29
Updated the svg checker so it doesn't post confirmation messages and updated CONTRIBUTING.md

Thomas-Boi:


amacado (approved):
Without a deep test this looks good to me! :) I just have the feeling that `cutenode/action-always-fail@v1.0.1` could be replaced without a vendor action. If I get it right, this action will just _fail_? This could probably done without any vendor component, or? 

Anyway: Thanks for improving and you have my approval! @Panquesito7 might wanna take a look before merging?

Thomas-Boi:
> 
> 
> Without a deep test this looks good to me! :) I just have the feeling that `cutenode/action-always-fail@v1.0.1` could be replaced without a vendor action. If I get it right, this action will just _fail_? This could probably done without any vendor component, or?

That's true. I originally wanted to have a clean way to fail the workflow and this action seems to be it. If this fails because of the action, we would have a message that says "This action intentionally fails etc..." rather than if we run a bogus command like "fake-command".

Issue #582, opened 2021-04-30
Icon request: Tensorflow


reevesba:
### Links
https://www.tensorflow.org/

tupaschoal:
This was added by #585 and is already live on https://devicon.dev/, so it can probably be closed.

amacado:
Thanks for the hint @tupaschoal!

PR #585, opened 2021-04-30
new icon: tensorflow (original, original-wordmark, plain, plain-wordmark, line, line-wordmark)

reevesba:
Issue #582


reevesba:
Looks like a couple of the svgs were corrupted during the resize. I will make another commit shortly.

Thomas-Boi:
Hey @reevesba,

Thank you for the icons 😃 

Unfortunately, there are a few issues that we'd need to address:
- The icons and svgs are not centered within its `viewbox`. Can you please change it so all SVG are centered?
- The `wordmark` versions of the icon are unfortunately not what we defined as wordmark. Here is what a `wordmark` version look like:
image

A possible fix would be removing the `wordmark` svgs or updating it so it contains the technology name as shown above. Note that you'll need to convert strokes to fill when working with texts since Icomoon (the icon making site) can only parse fills.

- I notice that the `original` and `plain` versions are exactly the same. In this case, you can remove one (I'd suggest the `plain` one) and remove its reference within the `devicon.json`. This will help keep the font and css file small.

Thank you again and I hope to see this icon in the repo soon.

reevesba:
Hi @Thomas-Boi, 

Thanks for the tips. The original did have color which was the only difference. I will go ahead and remove the plain versions and update devicon.json. I will also update the wordmark versions, convert strokes to fill and center the svgs in the viewbox.

Thomas-Boi:
Hey @reevesba ,

The icons look great! Thank you for your effort. We just need to get a feature into develop before we can merge your PR in. This should take a few more days and then, your icons will be accepted!

Thomas-Boi (approved):
So good news: we've decided to postpone the update for a few days. In short, we are trying to update the svgs in our repo so it's optimized and has an unique id within the whole repo. Since we will have to do this for the whole repo, we can accept your PR and update your svgs then rather than do it now.

In summary: Thank you for the icons @reevesba!

PR #597, opened 2021-05-04
Optimized SVG using SVGO

Thomas-Boi:
This PR contains the new (under development) workflow that uses SVGO to optimize our svg files (minify + prefixing ids). 

It also contains the optimized `svg` files. I did a few random checks and so far it look good. I'll manually go over each files in the next few days to ensure the `svgs` were optimized correctly. 

Fun fact: we have 530-ish icons! This is a both an interesting and a bit daunting fact considering I'll have to check them all 🙃. I'll split them up into small chunks and update everyone on the result. After that, I'll make this into a workflow, update the `Check SVG` bot and `CONTRIBUTING.md` (again 😆) to reflect this new workflow.

EDIT: Checked 30/530 icons. So far so good. Also, note that the optimizer will now remove the `width` and `height` attribute but keep the viewBox. This means that the user can now scale the svg in the webpage using a wrapper container rather then changing the svg styling itself.

EDIT: Checked 201/534 files. Still good so far.
EDIT: Checked 310/534. No problem detected yet.
EDIT: Checked 403/534. No problem seen. Found out that the `python-original.svg` is not centered vertically though.


Thomas-Boi:
Checked all 534 SVGs! All of them looks good and did not change from the previous. A couple changed its size because we remove the `width` and `height` but since we keep the `viewBox`, everything should be fine. In fact, this will allow the SVG to scale based on the container it's in, rather than being restricted at 128px by 128px.

I also found a couple duplicated svgs (original and plain are similar etc...). I think we can remove the extra and update the `devicon.json` to reflect it. Rebuilding the icons would be unnecessary since the font will still look the same as the svg. 

More features coming tomorrow.

Thomas-Boi:
Just tested the workflow. Everything works great and you can see the result [here](https://github.com/Thomas-Boi/devicon/pull/41/files?file-filters%5B%5D=.svg&file-filters%5B%5D=.yml)

I will now do some clean up and update the `check-svg` bot's check requirements.

Panquesito7:
> optimizer will now remove the `width` and `height` attribute but keep the viewBox

Wouldn't this make the bot create a warning? AFAIK, if you don't set `width` and `height` to `128` (or `128px`), it'll display a warning.

Thomas-Boi:
> Wouldn't this make the bot create a warning? AFAIK, if you don't set `width` and `height` to `128` (or `128px`), it'll display a warning.

Oh yeah, my plan is to remove that warning since our script would fix it for us.  You can see it being removed in the `check_svgs_on_pr.py`.

> Also: it's amazing how you have changed 10k+ lines.
Yeah, the optimizer did most the work. I still have to visually check the icon to ensure the optimization options won't mess anything up though. I think it took me at least 1-1.5 hour 😅

amacado (dismissed):
I just looked trough **529** icons. Crazy.
image

I could not see any fail in transitioning the icons. And it's odly satisfying to see icons get into shape:
image
image

Thanks for all the effort you had to put into this patch! If you could resolve the conflict we can merge and release it!

Thomas-Boi:
Hey @amacado , thank you very much for reviewing this. I did not expected you to go over the 500 SVGs so I apologize if you felt rushed for doing so 😅.

I'll fix the conflict asap.

amacado (approved):

Issue #607, opened 2021-05-14
Similar SVGs 


Thomas-Boi:
There are a few similar SVGs that can be removed to reduce the size of the project. Here are the ones that I've found (`og` means original and `pl` means plain).

- debian original and plain
-cake original and plain
- erlang original and plain
- gatsby og and pl
- jquery og and pain
- kraken js og and pl
~~- lua og and pl, og word mark and pl wordmark~~
- meteor og and pl
- perl og and pl
- phoenix og and pl
- python og is not centered
- spring og and pl, og word and pl word

I'll make a new PR removing the duplicates and update the `devicon.json` so it keeps the `original` version over the `plain` one.

Thomas-Boi:
I'll also update the `icomoon.json` and the usual font files to get the names to match. Some of the icons refers to the deleted SVGs so they need to be updated.

Thomas-Boi:
We've decided to not remove the extra files in case some people are using them. However, from now on, the maintainers will ensure that duplicated files are not submitted to the repo.

PR #103, opened 2018-02-16
Fix python wordmark icon

Kartstig:
- Relates to https://github.com/konpa/devicon/issues/93

PR #608, opened 2021-05-15
Removed duplicated svgs

Thomas-Boi:
This removes the duplicates listed in #607.

Notes:
- Spring, perl, and phoenix's plain versions were the same as their original but with the color stripped. The font versions also have the color stripped so we don't have to worry much about this.
- Lua actually don't have any duplicates, this has been crossed out in the issue
- Also centered `python-original.svg`
- Added `screenshots` folder to `.gitignore`. This is probably created due to one of the build steps


Thomas-Boi:
Since `python-original.svg` was changed, this is also our first official test of the `svg-optimizer` bot. Let's see how it do.

Thomas-Boi:
So it works great as expected. I'm also pleasantly surprised that the ids are prefixed properly when it's already prefixed.

Thomas-Boi:
Just updated the font files and the `icomoon.json` to the proper names. Also tested it and everything looks fine in the Icomoon website.

Thomas-Boi:
Just pushed a commit optimizing the `python-wordmark-original.svg` as seen in #103

.gitignore
```diff
@@ -4,4 +4,5 @@ node_modules
 geckodriver.log
 __pycache__
 *.pyc
-new_icons.png
\ No newline at end of file
+new_icons.png
+screenshots
```
Panquesito7, line 8: Missing an endline.
```suggestion
screenshots

```

amacado (commented):
I support the idea of reducing the overall size by removing duplicates. But since this is a breaking change I would like to discuss if it's worth it. On one hand we have the reduced file size on the other hand it could break existing applications when someone (like me) is using the icons in an automated way. How much space do we save when removing those icons?

Thomas-Boi:
> I support the idea of reducing the overall size by removing duplicates. But since this is a breaking change I would like to discuss if it's worth it. On one hand we have the reduced file size on the other hand it could break existing applications when someone (like me) is using the icons in an automated way. How much space do we save when removing those icons?

Not a lot actually. The files are between 1 kB to 4 kB so I'll use 2 kB per file. There are 12 files removed => `12 * 2 kB = 24 kB`. The benefit is small I admit and I have no strong feeling about the PR.

Thomas-Boi:
@amacado I think you are right. The file savings are not a lot but the potential backwards compatibility issue might be worth more. I'm cool if you want to close this PR. I'll make another one for the other fixes.

PR #634, opened 2021-05-27
new icon: rstudio (original, plain)

Thomas-Boi:
Taken from #116. Work was done by @Pmz64. I modified the plain version to get the original version.


amacado (approved):
thanks for picking this up @Thomas-Boi

PR #640, opened 2021-05-27
new icon: matlab (line)

Thomas-Boi:
Taken from #116. Work was done by @Pmz64.

### DIFF
--- a/devicon.json
+++ b/devicon.json
@@ -2339,10 +2339,12 @@
         "versions": {
           "svg": [
             "original",
-            "plain"
+            "plain",
+            "line"
           ],
           "font": [
-            "plain"
+            "plain",
+            "line"
           ]
         },
         "color": "#6dd0c7",
--- a/icons/matlab/matlab-line.svg
+++ b/icons/matlab/matlab-line.svg
@@ -1,0 +1,66 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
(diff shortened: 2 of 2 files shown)


amacado:
The problem is within the `find_new_icons` method https://github.com/devicons/devicon/blob/7a4ca8aa871d6dca81691e018d31eed89cb70a76/.github/scripts/build_assets/filehandler.py#L9

Because `matlab` (plain version) is already in our icon set (`if is_not_in_icomoon_json(icon, icomoon_json)`) the line version is not detected as _new_ icon, which results in a failed check.

Thomas-Boi:
@amacado you are right. 

There's also another issue: the build bot won't build this icon either because the folder `matlab` is already built in `icomoon.json`. See `filehandler.py`'s `find_new_icons()` for more details.

I can see two options:
- Build this file manually (quickest)
- Change the build script so whenever it runs, it builds EVERY single SVGs listed as `font` versions. 

I'm leaning towards option 1. However, we might need to address icon updates in the future as well.

amacado:
@Thomas-Boi would you work on a fix? I think this case could occur more often in the future.. Maybe we should create a issue for this.

Panquesito7 (approved):
Thanks! The icon itself and the `devicon.json` files look good. Good work. 👍 🎉

Thomas-Boi:
Here's the [issue](https://github.com/devicons/devicon/issues/656) related to this:

Thomas-Boi:
I'll merge this in so I can test the new build script. Besides, this is good anyway

PR #641, opened 2021-05-27
Remove new_icons from gitignore

Thomas-Boi:
This is preventing the `peek-comment` bot from functioning properly.

@amacado or @Panquesito7 please merge this PR first then update the develop branch of each other PR. Currently, those branches cannot finish the peek task and the fix in here will address that. Then, all you have to do is re-apply the `bot:peek` label.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,8 @@ node_modules
 geckodriver.log
 __pycache__
 *.pyc
+
+
+# DO NOT ADD THE FOLLOWING FILES TO THE GITIGNORE
+# new_icons.png - this is used by peek bot
+# screenshots/ - this is used by build bot
--- a/.gitignore
+++ b/.gitignore
@@ -4,4 +4,3 @@ node_modules
 geckodriver.log
 __pycache__
 *.pyc
-new_icons.png


amacado (commented):
Can you explain why this prevents the bot from working properly?
I took a look at the actions log but I don't understand why this patch is the solution. Since we do not `add/commit` the screenshots they can be ignored, or? The only task is to extract them from a zip (which is done) and then uploading them to imgur (which failes with 403).

image

Since the action succeeds in other cases like https://github.com/devicons/devicon/runs/2681459934?check_suite_focus=true I don't think it's a problem of the gitignore. But I might be wrong, so I hope @Thomas-Boi can explain and enlighten me! :)

Thomas-Boi:
@amacado 
I think you are right. I was confused by the error and didn't remember that the peek workflow doesn't commit anything. I'll be closing this branch since this is not a fix as I thought

Thomas-Boi:
I just retried applying the `bot:peek` label and some of the PR now works. I don't what's happening. I think the only thing we can do is retry a few times and see what happens

PR #166, opened 2019-05-18
Icons: Karma, Jest, Nest, CircleCI, Codecov

rjlopezdev:
# Description
This PR add modern platform/framework icons:

* [Karma](https://karma-runner.github.io/latest/index.html) (original, plain)
* [Jest](https://github.com/facebook/jest) (plain)
* [Nest](https://nestjs.com/) (original-wordmark, plain, plain-wordmark)
* [CircleCI](https://circleci.com/) (plain, plain-wordmark)
* [Codecov](https://codecov.io/) (plain)

PR #646, opened 2021-05-27
new icon: karma (original, plain)

Thomas-Boi:
Taken from #166.


Thomas-Boi:
I'm looking through the logs and we have an error of 403 (Forbidden Error). I think this might be an API issue.

On the other hand, this works just fine a few hours ago. I don't understand how we are getting denied

Thomas-Boi:
Figured out why: Imgur changed their URL. 

Here's what we have in the `upload-to-imgur` action:
```
  const axiosConfig = {
    method: "post",
    url: "https://api.imgur.com/3/image",
    headers: {
      "Authorization": `Client-ID ${clientId}`
    },
```

Here's the new url:
image


I'll just update the other repo and we should be fine

amacado:
`bot:peek` is still running the old version, since it's not yet published in the master branch:
image

but re-running the action several times works too ;-)

amacado (approved):

PR #649, opened 2021-05-27
new icon: codecov (plain)

Thomas-Boi:
Taken from #166.


amacado (changes requested):
I think this icon needs some rework. If I compare it to the original there is a difference in the way it looks:

image

taken from https://about.codecov.io/

compared to:
image

amacado (changes requested):
Could you consider to align the icon in the center? But thanks anyway for recreating this icon!

image

amacado (approved):

PR #651, opened 2021-05-28
update public-upload-to-imgur action v2.2.2

amacado:
This will hopefully fix https://github.com/devicons/devicon/pull/646#issuecomment-849959829


Thomas-Boi (approved):
Good catch 👍

Issue #614, opened 2021-05-18
Issue with Optimize-Bot on forked repo PR


Thomas-Boi:
Hello guys, 

I discovered an issue with the `optimize-bot` as seen in #610. The `checkout` and `commit` action acts differently when the PR is from a forked repo.

The [`commit` action](https://github.com/stefanzweifel/git-auto-commit-action) calls for the `checkout` action to checkout the `github.head_ref` when the trigger is `pull_request`. Unfortunately, this cause the action to checkout the PR branch's name but look for it in the `devicons/devicon` repo => does not exist.

Removing the `github.head_ref` would cause the `commit` action to fail entirely but everything else works.

I'd like to ask you guys what you think of the following potential solutions:
- Changing the `commit` action into a `create-pr` action. This is similar to what we already have in the `build-fonts` action.
  - The target will be the PR branch:
    - Pro: Don't need to close the branch and keep everything in the same PR
    - Con: Might need permission of the owner to merge. If the owner has abandoned the PR, this can be problematic. However, if the owner has abandoned the PR, we'll have to take it over nonetheless and can PR into our own branch.
  - The target will be `develop`:
    - Pro: We can merge it in whenever.
    - Con: We'll have to close the old PR manually. PR will also be cluttered

- Keep the `commit` action but rather than using `files-changed` action, manually search for files added based on the PR title. 
  - Pro: Easy to do. `peek-bot` is using this.
  - Con: Hacky. Depends on PR title to be correct. Will need a wrapper Python script to run the npm script but that's not too bad.

- Run the `optimize` script another way: on merge, when building, etc...
  - Pro: Batch things together and save time
  - Con: need more work.

I'm thinking of doing option 2. We haven't have an issue with `peek-bot` yet. While it's a bit hacky, it works well and I won't have to do much to add this functionality.

What do you guys think?

Thomas-Boi:
Ok, after doing some work, I realized that option 2 does not work. I misunderstood/did not comprehend the docs for the `commit` action. Turns out that we will need to make the contributor enables the Action on their forked repo.  See [this](https://github.com/stefanzweifel/git-auto-commit-action#using-the-action-in-forks-from-public-repositories) for more details.

I'm thinking that optimizing the SVGs all at once when we do a release (option 3) might be better idea. Either that or I look into the `PR` action (option 1).

Thomas-Boi:
[...]

PR #652, opened 2021-05-28
new icon: magento (original, original-wordmark, line)

Thomas-Boi:
Taken from #193. The `plain` versions are the same as the `original` so they were removed.


Thomas-Boi:
The `post comment` bot is trying to upload 3 images. However, the API would fail at the 3rd image. Action result [here](https://github.com/devicons/devicon/runs/2696629605?check_suite_focus=true)
image

I think this is more of an issue with their API than us. I'll contact them and see if they can offer any solution.

Panquesito7 (commented):
Maybe you should consider optimizing these SVG's.

Thomas-Boi:
I'll check what's up with the line version

Panquesito7 (commented):
Are all these SVG's optimized with some tool? It seems they aren't. 🤔

Thomas-Boi:
They aren't optimized yet since we are using a new optimization step. See this [issue](https://github.com/devicons/devicon/issues/614)

Panquesito7 (approved):
Looks good. Great work, @Thomas-Boi! 👍 🎉

PR #655, opened 2021-05-28
new icon: vuestorefront (original, plain)

Thomas-Boi:
Taken from #193. The wordmark version was removed because the words don't match the wordmark on their website.


Thomas-Boi:
I'll update this font

PR #657, opened 2021-05-28
Peek bot won't check whether icon is in icomoon

Thomas-Boi:
This will address the issue in #640 where `peek:bot` will not run since `matlab` is already built.

Test locally works. As long as the title matches the format the `peek:bot` uses, everything is fine.
image


Panquesito7 (approved):
I don't know too much about Python, but if it works, then this looks good to me. 👍

Thomas-Boi:
Pretty much, the peek bot will read the content of the `devicon.json`, check the PR title, and run its task. 

Before, after it reads the `devicon.json`, the peek bot also compares it with the `icomoon.json`. This means folder that were built were not checked. I just remove that step basically.

PR #660, opened 2021-05-31
Fix svg checker double posting comment

Thomas-Boi:
Currently, the SVG checker bot will post two messages whenever it finds an error with the SVGs. This PR removes the trigger for the second message. Now, it will only shows up in case of failures.


amacado (approved):
Finally! :) This bug was annoying 😆

Issue #656, opened 2021-05-28
Build bot currently not building updated icons


Thomas-Boi:
**Problem**
Adding a new SVG to a folder that has been built will not make the new SVG to be built. This also applies to updating an SVG within a built folder 

**Current Situation**
The `build bot` only searches whether an icon folder has been added to the `icomoon.json`. If it has, that entire folder and its content, will not be built. This was done so searching through the `icomoon.json` is faster than searching every icon.

In the screenshot below, I'm looping through the `icomoon.json` and if I found a name (ex. apple), I'm finished.
![image](https://user-images.githubusercontent.com/43018778/120024586-4817e780-bfa4-11eb-9acc-5eedec79d17d.png)

**Possible Fixes**
I can think of four ways of doing this:
1. Manually update the fonts files and css files whenever this happens. 
  - **Pro**: This is the quickest in the short term
  - **Cons**: Will be tedious in the long run. Not recommended.
 
2. Make the `build bot` rebuilds the entire repo every time it's run. 
  - **Pro**: This will ensure updated and added files are rebuilt properly as long as the `devicon.json` is up to date
  - **Cons**: Might take a long time to finish building. Time will get worse in the future. 

3. Add an extra property to the `devicon.json` that tells the bot whether a folder has been updated. The bot, when building, will also checks for this property and add the "updated" icon to the build list.
  - **Pro**: Fix the issue and run faster than option 2.
  - **Cons**: Extra property in the `devicon.json`. However, this property is optional and only need to be added when an icon needs to be re-built.

Example how option 3 will work. We need to update the `codecov` folder. Maybe old icons got updated. Maybe we add more versions to the folder. We will add the `updated` property to the `codecov` object in `devicon.json`.
```
    {
        "name": "codecov",
        ...,
        "updated": true
    },
```

The bot will parse the file and add them to the build list. Afterwards, it will rewrite the `devicon.json` and remove that attribute.

4. Rather than comparing the `devicon.json` and `icomoon.json`, the bot will check the PR history since last build and extract the icon name from the PR titles. 
  - **Pro**: This is already being done in `get_release_message` bot. Since all icon PR are tagged with `feature:icon`, finding icon PR would be easy.  Extracting the icon name is easy as well since `peek-bot` already does this.
  - **Cons**: Require all icon PR to be in the same format and be tagged with `feature:icon`.

What do you guys think?

Thomas-Boi:
[...]

PR #666, opened 2021-06-08
Build bot now build new SVGs in folder that were already built

Thomas-Boi:
Hello,

This PR contains the following changes:
- Build bot now uses past PRs to get which folder were added/updated. This means we now have to enforce PR title and PR labelling for icon requests.
- Build bot will check the `icomoon.json` and remove any icons that are listed in the above step. Ex: `matlab` has a new icon version `line`. The bot will remove the entry for all `matlab` icons in the icomoon.json`.
  - Note: the entire `matlab` folder will be reuploaded after this step. Thus, even if we removed them here, the icons will still be present in the icomoon and CSS file.

I've tested the bot locally to ensure it works. Here are the test results:

Before deletion (note the number of icons):
before_update

After deletion (note the number of icons):
after_update

Comparing after and before deletion. Note that the `matlab-plain` icon is not in the first set.
matlab_was_deleted

Bot logging:
logging

Reuploading the `matlab` folder. Recall that `matlab-line` was the new icon while `matlab-plain` did not change. However, since we remove `matlab-plain` in the `icomoon.json` update step, we will reupload them here:
full_run_log

All `matlab-icons` are added properly after the build step:
reupload_works

Let me know if you guys have any questions 😄

UPDATED: This was added due to #640 and #656
UPDATED 2: After this and deciding whether Excel should be accepted, I think we can make a new release.


amacado (approved):
Let's give it a try! :) Thanks for all the effort @Thomas-Boi

PR #691, opened 2021-06-23
new icon: Qt (original)

tupaschoal:
Tried to do it with the best of my knowledge of SVG (which is not a lot) and following the [contributing](https://github.com/devicons/devicon/blob/master/CONTRIBUTING.md) document.

I use the SVG downloaded directly from [Qt's website](https://brand.qt.io/design). It's pretty simple, so I only added the `original`:
image

This closes #190

PR #692, opened 2021-06-24
Only build script will need token API

Thomas-Boi:
This fixes the issue seen in #691 with the peek bot.

Issue:
- The script failed because the script requires a `token` command line arguments. However, peek bot does no need them.

Reason:
- The build bot and peek bot share most of the same command line arguments. One of them, the `token`, was recently added and was meant for build bot only. However, I forgot to enforce that by accident.
- The fix uses an existing check to ensure only the build bot requires the token


Panquesito7 (approved):
Thanks. 👍

PR #700, opened 2021-06-26
new icon: GCC (original)

tupaschoal:
I used the SVG downloaded from [Wikipedia](https://commons.wikimedia.org/wiki/File:GNU_Compiler_Collection_logo.svg) and tinkered it to be added as the original:
image

I also did a plain version, but I didn't know how to add it, since the for it the predominant color is black, rather than the beigeish for the original one (which is the `#ffcfab` used on the JSON). Pleas advise on how to proceed.

image


Thomas-Boi:
For the plain SVG, you can add it in this black color or make it all beige. As long as the plain SVG only consists of one color, that is fine. Note that the plain ICON (the font one) will only be black since we strip colors when we make it into a font. The SVG though can have any color (see [this](https://github.com/devicons/devicon/blob/master/icons/ruby/ruby-plain.svg) example) as long as it's only one.

tupaschoal:
Alright, thanks @Thomas-Boi. I think that this is done then, hopefully

Thomas-Boi (approved):
Everything looks good @tupaschoal!

PR #701, opened 2021-06-26
Fix plain icon example on CONTRIBUTING

tupaschoal:
The file name example is wrong for the plain version, as shown below:
image

This PR just fixes that.


Thomas-Boi (approved):
Thank you for spotting this! 😄

PR #704, opened 2021-06-28
Add a color picker in the header

### LINKED ISSUE #698: Feature Request: Background in the website should be able to change its color
### Problem 
As seen in #685, some icons will not show up properly over the default background. 

### Suggested Feature
We should have a color picker function that changes the background of the icons/SVGs section on the left side (the boxes). 
image

By default, perhaps we can use the checker-pallette. The color picker will then allow the user to change the background for more freedom if the checker-board is not enough.
image

### Why we should have this feature
- None beside fixing the stated problem

Thomas-Boi:
**Feature:**
- This add a color picker functionality to the background of the font and SVGs in the header (the aside technically).
image

**Notes:**
- Each font and SVG section has its own color. When the user select a new icon, the color will reset to the default color, which is our page color.
- I thought of using a checkerboard pattern but decided against it. It doesn't look good for some icons:
image

This will closes #698.


amacado (approved):
The UI could be prettier I guess, but it will do it's job ;-) Thank you @Thomas-Boi

Thomas-Boi:
@amacado I can change the color picker UI, just give me some suggestions :)

There are a few libraries out there or we can style it differently. Maybe a button with an icon that will change color depending on the color input? 

image

PR #712, opened 2021-07-03
new icon: jamstack (original, original-wordmark, plain-wordmark)

tupaschoal:
`Original`:
svg_2
`Original-wordmark`:
jamstack-original-wordmark
`Plain-wordmark`
jamstack-plain-wordmark

I was a bit confused on whether I should define those as `line` or not (or at least alias them as `line-...` Let me know what you think


Thomas-Boi:
I think this is fine. An informal rule that I follow is if it's unclear, the priority is `original` > `plain` > `line`. This is because `original` and `plain` are more common than `line` => people are more likely to search up those versions.

Thomas-Boi (approved):
The icons look great! Thank you for your work 😃

PR #725, opened 2021-07-07
new icon: filezilla (original)

MichielPater:


Panquesito7 (dismissed):
I checked out their website, and it seems you're missing an `original-wordmark` version. Please add it here in SVG format, too.

image

MichielPater:
@Panquesito7 Thank you for your changes!

I've seen the wordmark logo, but I'm not sure how to convert it to SVG format without losing precision. Do you have any suggestions?

Thomas-Boi (changes requested):
We're very close. This need to solve one final hiccup. As you can see in the bot's images, the `original` version is black. This is because we strip colors from the logo. In the image, I can see that the `f` is white => remove the colors and its black, same as the red background.

Suggestion:
If you have Inkscape, you can click on the SVG object -> decompose it into smaller objects -> click on the white `f` -> delete it. Then, replace the `original` entry in `font` with `plain`.

This *should* leave the red background and a transparent `f`. Then, save it as `filezilla-plain.svg` and resubmit. If the `f` is not clean (ex. jagged edges when removed), we might have to find another way. Props to you if you can do the same thing to `original-wordmark` and make it `plain-wordmark`.

MichielPater:
@Thomas-Boi Updated the icons :)

devicon.json
```diff
@@ -1300,6 +1300,22 @@
...
+        "name": "filezilla",
+        "tags": [
+            "ftp"
+        ],
+        "versions": {
+            "svg": [
+                "plain", "plain-wordmark"
+            ],
+            "font": [
+                "plain"
```
Thomas-Boi, line 1314: This looks great now! I also notice that you changes the `wordmark` version as well, which will make it font-friendly. I think we can add this in too.
```suggestion
                "plain",
                "plain-wordmark"
```

Thomas-Boi (approved):
@MichielPater fantastic work! `Filezilla` looks great now. Thank you for your time and welcome to the repo!

PR #737, opened 2021-07-08
Latest os

theblobscp:
Updated workflows to use the latest verison of the os

### DIFF
--- a/.github/workflows/build_icons.yml
+++ b/.github/workflows/build_icons.yml
@@ -3,7 +3,7 @@ on: workflow_dispatch
 jobs:
   build:
     name: Get Fonts From Icomoon
-    runs-on: windows-2019
+    runs-on: windows-latest
     steps:
       - uses: actions/checkout@v2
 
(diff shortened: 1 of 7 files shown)


Panquesito7 (approved):
Looks good. Thank you for your contribution! 😄👍

theblobscp:
Thanks! But also sorry for triggering like 10 different workflows on my pr

Panquesito7:
> sorry for triggering like 10 different workflows

It's alright. 🙂

Thomas-Boi:
Hi @theblobscp,

Unfortunately, I don't think we can accept this PR. The current OS versions are locked for compatibility reasons and ensures that the build scripts work in all cases (a bit like a Docker container). 

For example, `windows-latest` will always have the latest change, which can be `windows-2021` or `windows-2030` in the future. However, `windows-2030` might function differently than `windows-2019` and could break the workflow. We already know `windows-2019` works and as long as GitHub provides the OS image, the workflow will also work. Thus, rather than forcing us to change every time a new OS comes out, we can keep using the current script until it's no longer supported.

I'll close this PR for now but I'm willing to reopen if you make good points for changing to latest OSes.

theblobscp:
Oh ok. Well, let's just hope windows has good backwards compatibility then because I can't script at all.

PR #755, opened 2021-07-17
Add a CSS library (foundation/bootstrap) to devicon.dev demo website

theblobscp:
Most websites use either bootstrap or foundation.  I added both as a html comment so you can pick which one would be best.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -32,7 +32,14 @@
   <meta name="msapplication-TileColor" content="#da532c">
   <meta name="msapplication-TileImage" content="./logos/mstile-144x144.png">
   <meta name="theme-color" content="#ffffff">
-
+<!-- pick this one if you want bootstrap
+<link href="https://cdn.jsdelivr.net/npm/bootstrap@5.0.2/dist/css/bootstrap.min.css" rel="stylesheet" integrity="sha384-EVSTQN3/azprG1Anm3QDgpJLIm9Nao0Yz1ztcQTwFspd3yD65VohhpuuCOmLASjC" crossorigin="anonymous">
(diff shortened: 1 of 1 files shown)


amacado (changes requested):
@theblobscp thanks for the hint, but honestly I don't see how this would help the project? Since our demo website (devicon.dev) is up and running (sure, there's always room for improvement), how does adding a css framework - which is not used yet - helps? 

When you have a particular idea how to improve the website I'm curious to hear about it, but adding frameworks alone just increases the load time without any benefit. Maybe I'm on a false track, so please prove me wrong! 😃

theblobscp:
Mostly due to the fact bootstrap has prebuilt classes (example d-flex), which saves the time of writing the same lines of code over and over again

amacado:
> Mostly due to the fact bootstrap has prebuilt classes (example d-flex), which saves the time of writing the same lines of code over and over again

while that is generally true, can you point out where exactly you see the improvement in our code?

theblobscp:
I would use the vercel thing to show differences but I was never able to get it to work

Thomas-Boi:
Hey @theblobscp,

For our website, we've styled it without Bootstrap. There are also some angular-specific styling in there and I do not want to mess that up. 

Furthermore, while it's true that bootstrap is popular, it doesn't mean that every project, or our website, need to use it. For the devicon site, we've never use it and I don't think we need to in the foreseeable future. In this case, it's the rule of "if it's not broken, don't fix it".

I know that in programming, there seems to be things that EVERYBODY do. However, the repo is set up in a way that suit us and we have no need to change it. If you make a compelling argument for why we need bootstrap (say, a new layout), I'm willing to change my mind. For now, I'll close the PR.

PR #765, opened 2021-07-21
Closes #693 - Fix issue with vuestorefront having wrong folder name 

### LINKED ISSUE #693: Bug Report: svg file of "vuestorefront" not exist.
`vuestorefront` SVG is not exist, but `devicons.json` is written about it

スクリーンショット 2021-06-24 20 30 24

https://github.com/devicons/devicon/blob/9f4f5cdb393299a81125eb5127929ea7bfe42889/devicon.json#L3822-L3837

In fact, svg type `original` and `plain` is not exist 🙌

Thomas-Boi:
This should fixes #693.

Git doesn't track capital-sensitivity in file paths, thus a simple file name changes locally doesn't fix it.  This PR basically removes the folder in one commit and restore it in a different name in another, hopefully causes git to track it.


amacado (approved):
>Git doesn't track capital-sensitivity in file paths

yep, I had this problem in other projects as well.. You came up with a solution, I will use it in some other cases 👍🏻

Panquesito7 (approved):
Thank you for pointing this out and for the fix. 🙂

Thomas-Boi:
I'll merge this in. The icons are still the same so this won't show up in the build release

PR #775, opened 2021-07-26
Use latest python verison for workflows

theblobscp:


Panquesito7 (approved):

Thomas-Boi:
Hi @theblobscp,

While I appreciate the thought, I would like to keep the same Python version for as long as possible. This will ensure that the script will work as long as the version 3.8 is offered by GitHub (think Docker containers). I'll only update the version if the company is dropping its support, the script stopped working due to Selenium needing a new version, or there's a security reason (like the current npm update that I'm working on).

I'll close this PR for now. However, if you have a good reason to move to Python 3.9, I'll reopen this.

PR #791, opened 2021-08-09
Remove aliases in the OpenGL icon as they're not needed

Panquesito7:
Things added/changed:

- Remove aliases in the OpenGL icon as they're not needed.


Thomas-Boi (approved):
This looks good! Thanks for the help @Panquesito7 👍

PR #792, opened 2021-08-09
Fix the wrong `viewBox` in the `towergit-plain-wordmark.svg` file

Panquesito7:
Things added/changed:

- Fix the wrong `viewBox` in the `towergit-plain-wordmark.svg` file.


Thomas-Boi (approved):
Thanks for the fix. I'll merge the changes now

PR #807, opened 2021-08-13
Feature: check-svg-bot upgrade

Thomas-Boi:
Hello guys,

This PR contains upgrades to the `check-svg` bot. The bot will now:
- Not check for `<style>` tags in the SVG. Since the optimizer bot will prefix the classes with the file's name, the class name will be unique from now on => avoid issues of clashing class names from using multiple SVGs inline within HTML.
- ~~Check for `stroke-width` in the property and styles of the elements. After some testing, I found that Icomoon will warn us about strokes based on the presence of the `stroke-width` property/style.~~

Test: N/A

EDIT: I've removed the `stroke-width` check. Having it in would require refactoring the script to almost similar as `peek-bot`. I think having `peek-bot` checking for strokes should be enough. Let me know what you guys think. If you think having `check-svg` warn them is better (since it's earlier), let me know.


amacado (approved):

Panquesito7 (approved):
Thanks, @Thomas-Boi.

PR #813, opened 2021-08-16
Feature/website cache busting

MichielPater:
Added cache busting, so that cache will be cleared on client side whenever devicon.json or SVG files are updated.


Thomas-Boi (dismissed):
Michiel and I have discussed this with each other and the changes look good. I also tested the URL to fetch `devicon.json` manually and the result is still the same. 

After all, the `?` adds query params but since the API don't accept params for this, nothing happens.

Panquesito7 (approved):
For me, it looks good. I'll better let @amacado check if this PR's fine, though. Thank you for your contribution, though! 😄👍

docs/index.html
```diff
@@ -135,7 +135,7 @@ <h3 class="cbp-ig-title">{{icon.name}}</h3>
   <script src="//ajax.googleapis.com/ajax/libs/angularjs/1.2.19/angular.min.js"></script>
   <script src="//ajax.googleapis.com/ajax/libs/angularjs/1.2.19/angular-sanitize.min.js"></script>
   <script src="//ajax.googleapis.com/ajax/libs/angularjs/1.2.19/angular-animate.min.js"></script>
-  <script src="assets/js/script.js"></script>
+  <script src="assets/js/script.js?2.13.0"></script>
```
amacado, line 138: Having a static value of `?2.13.0` will probably not achieve the required result, or?

MichielPater: That's only for the script file, to make it update with the cache busting feature. From this point, the script only needs to be cache busted whenever that file changes, but wouldn't as long as it stays the same. The cache busting is done within the file.

Thomas-Boi: If the file is in angular (which it is), can't we just use variables defined in `$scope`? That way we never have update the version manually.

MichielPater: That is true. However, I don't think we can use the variables we created within the script file itself. But we could use others, or move the definition outside of the file.

Or maybe there is even caching set outside of these files, like in a configuration or the HTML file. The PR was intended to be a small change though.

Thomas-Boi: Perhaps we can embed angular code in? Since angular files were listed before the `script.js`, perhaps we can do something like:
```
<script src='script.js?{{ (new Date()).getTime() }}' />
```

I'm not 100% sure this is feasible with angular but I'd think it's possible. As for caching setting for the file itself, we'll have to check the GitHub pages doc

MichielPater: We could replace this <script> tag with the following?

```
<script>
	document.write("<script src='assets/js/script.js?" + Date.now() + "'><\/script>");
</script>
```

Thomas-Boi:
Just tested @MichielPater's solution for cache-busting JS and it works well
image

The only thing that's needed is putting the script tag for the `document.write` in its separate `<script>

Panquesito7 (approved):

Panquesito7 (dismissed):

Thomas-Boi (approved):
This looks good! Thanks for the help 😄 @MichielPater

Thomas-Boi (approved):

PR #815, opened 2021-08-18
Fix syntax with the issue templates

### LINKED ISSUE #814: [BUG] The Icon Request Template is broken
### Bug description

The Icon Request Template doesn't show up when I click on `New Issue`.
image

When I check the file, GitHub said there's an issue:
image

Remove the validations as suggested by the code seems to be a solution.

### Additional information

We might need to push a patch update so people can use this issue. We can do it at the same time as the website upgrade.

Panquesito7:
Things added/changed:

- Fixed the issue templates having some syntax errors.
  - Sorry for removing the PR template, but this is just a simple bug fix, so I don't think it needed the template. 😅
  - Closes #814.


Thomas-Boi (approved):
This looks good 👍. Thanks @Panquesito7

PR #827, opened 2021-08-27
update icon: msdos (original, plain, line)

MichielPater:
## This PR adds/fixes

- msdos icon: removed plain version from font for now as discussed in previous PR.


Thomas-Boi:
Hey @MichielPater,

I'm guessing that the `plain` version doesn't fit your standards? 
image

I think it doesn't look as good as the `line` version either. Perhaps let's remove the `plain` file as well along with your current change?

MichielPater:
Oh I thought the SVG looked good on the website, just the font looked different!

Thomas-Boi (approved):
I see what you mean. In that case, I'll just update your PR so our bot can remove the plain version in the next build.

PR #848, opened 2021-09-09
New Feature: Run Repository Dispatch for the React Repo on a NPM Release

maltejur:
# **New feature section**

- [x] PR name matches the format *New Feature: a brief description of feature*

## This PR adds/fixes

This PR adds another step in the `npm_publish.yml` workflow to run a [repository dispatch](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#repository_dispatch) to the react-devicons repo.

## Notes

To make this work, the environment variable `REACT_REPO_ACCESS_TOKEN` needs to be set to give the workflow access to the react repo. See more here: https://github.com/peter-evans/repository-dispatch

### DIFF
--- a/.github/workflows/npm_publish.yml
+++ b/.github/workflows/npm_publish.yml
@@ -21,3 +21,8 @@ jobs:
       - run: npm publish
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_AUTOMATION_TOKEN }}
+      - name: Dispatch React Release
+        uses: peter-evans/repository-dispatch@v1
+        with:
+          token: ${{ secrets.REACT_REPO_ACCESS_TOKEN }}
+          event-type: publish


Panquesito7 (approved):

Thomas-Boi:
Hey @amacado, can you check the note of this PR and add the environmental variable to the repo please?

.github/workflows/npm_publish.yml
```diff
@@ -21,3 +21,8 @@ jobs:
       - run: npm publish
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_AUTOMATION_TOKEN }}
+      - name: Dispatch React Release
+        uses: peter-evans/repository-dispatch@v1
+        with:
+          token: ${{ secrets.REACT_REPO_ACCESS_TOKEN }}
+          event-type: publish
```
amacado, line 28: If I get the description of https://github.com/peter-evans/repository-dispatch#action-inputs right, this will dispatch the "publish" event for this repository. I can't see any relation to the [react-devicons](https://github.com/devicons/react-devicons) repository. The documentation says if  `repository` property is empty it will use the current repository. Which is "devicons/devicon" in this case, not the react repo. I might be wrong, so I would be glad if you could explain this. 

I also see a problem in the ACCESS_TOKEN. As far as i know it's not possible to limit an access token to a specific repository. There are "deploy keys" which are repo-specific, but access tokens beeing generated on a user base (afaik). Maybe there is another way of solving the addressed issue?

amacado (changes requested):

.github/workflows/npm_publish.yml
maltejur: > If I get the description of https://github.com/peter-evans/repository-dispatch#action-inputs right, this will dispatch the "publish" event for this repository. I can't see any relation to the react-devicons repository. The documentation says if repository property is empty it will use the current repository. Which is "devicons/devicon" in this case, not the react repo. I might be wrong, so I would be glad if you could explain this.

You are totally right I somehow missed to add that.

> I also see a problem in the ACCESS_TOKEN. As far as i know it's not possible to limit an access token to a specific repository. There are "deploy keys" which are repo-specific, but access tokens beeing generated on a user base (afaik). Maybe there is another way of solving the addressed issue?

Yes this is a very valid concern. Afaik deploy keys can't trigger workflows, they are only used for cloning private repositories in a deployment process (more [here](https://github.blog/2015-06-16-read-only-deploy-keys/)). Since the whole release process in the main repository is more or less triggered manually, maybe we should just also trigger the `react-devicons` release script manually in this whole release process. I think that would also be the least error-prone way to do the release and it shouldn't be that much more work (just one button click in the `react-devicons` repo).

PR #902, opened 2021-11-03
fix: endlines in the CMake icons

Panquesito7:
Things added/changed:

- Fix endlines in the CMake icons.


Thomas-Boi:
Hey @Panquesito7, what is the goal of this fix? I'm fine with it but the minifier will minimize and append a `\n` anyway. Thus, there's no point to removing the `\n` manually.

Panquesito7:
> Hey @Panquesito7, what is the goal of this fix? I'm fine with it but the minifier will minimize and append a `\n` anyway. Thus, there's no point to removing the `\n` manually.
> 
> I can still merge it but I think you don't have to worry about this in the future 😄

The goal of this is to prevent issues in other people's IDEs. Adding an endline will ensure no problems. 🙂

Thomas-Boi:
Ah I see, I'll merge it then.

I'll also look at trying to add a `\n` script wise. Perhaps people make some script to do it now that more people use GitHub Actions

Thomas-Boi (approved):

PR #914, opened 2021-11-16
fix: make `tags` an array for `elixir` in `devicon.json`

maltejur:
## This PR adds/fixes

This PR makes the `tags` field in `devicon.json` for `elixir` an array, like it is for every other icon.


Panquesito7 (approved):
LGTM 🚀 Thank you for your contribution! 😄👍

PR #1058, opened 2022-03-13
New Feature: Changed to select all rather before take a pic

Thomas-Boi:
# **New feature section**

- [x] PR name matches the format *New Feature: a brief description of feature*

## This PR adds/fixes

Fixed the issue caused by Selenium being unable to click on an element in the webpage.

## Notes

Don't need to test the code. It's what we used before but later changed due to aesthetic reason. The broken code would only select the new icon to take a pic. This one would select everything so we'd have to look at every ourselves.

EDIT: There's only 1 line of code that was changed, which is line 95. IDK why the whole file was changed.


amacado (approved):
>EDIT: There's only 1 line of code that was changed, which is line 95. IDK why the whole file was changed.

It's because of changes in whitespaces, which can be hidden using an option while comparing the files:
image

Thomas-Boi:
Cheers, thanks @amacado 👍

PR #1169, opened 2022-04-30
Convert Style Elements to Attributes

maltejur:
In the process of trying to release the newest version of the react devicons, I got an error caused by the usage of style elements. The standards state that those style elements shouldn't be used

> Each .svg must use the fill attribute instead of using classes for colors. This is to prevent class name clashing when using inline SVG. See https://github.com/devicons/devicon/issues/407 for more details.

so this pull request fixes all the icons which have style elements.

----

@thehedgefrog since all those icons were from you and I saw that you used Illustrator to create them, when you export the icon there is should be an option to include the css properties as `Presentation Attributes`. In the future you should use that (`Style Attributes` should also be fine).

image


thehedgefrog:
Oh my bad! I had missed that (and am still quite new at SVGs), sorry!

I'm happy to redo them if you'd prefer.

maltejur:
No problem

> I'm happy to redo them if you'd prefer.

Thanks but there is no need for that, i have already cleaned up all icons in this PR

amacado (approved):

Issue #1125, opened 2022-04-01
[ICON REQUEST]: JetBrains Products


Kek5chen:
### I have searched through the issues and didn't find my problem.

- [X] Confirm

### About the icon

JetBrains has great products and provides them with great care.

### Links and sources

Here you can see the SVGs of all their developer tools:
https://www.jetbrains.com/products/

I've done a bit of own work but since contributing to this repository is very strict and I am afraid to make too many mistakes I forked it, placed some icons and pushed it to my fork. You can see the changes here:
https://github.com/Kek5chen/devicon/commit/0a06ff51f0a88db2549bd71d63a5c5a2f2dc39d8

Icons included:
[CLion](https://github.com/Kek5chen/devicon/blob/0a06ff51f0a88db2549bd71d63a5c5a2f2dc39d8/icons/clion/clion-original.svg)
[DataGrip](https://github.com/Kek5chen/devicon/blob/0a06ff51f0a88db2549bd71d63a5c5a2f2dc39d8/icons/datagrip/datagrip-original.svg)
[IntelliJ](https://github.com/Kek5chen/devicon/blob/0a06ff51f0a88db2549bd71d63a5c5a2f2dc39d8/icons/intellij/intellij-original.svg)
[PhpStorm](https://github.com/Kek5chen/devicon/blob/0a06ff51f0a88db2549bd71d63a5c5a2f2dc39d8/icons/phpstorm/phpstorm-original.svg)
[PyCharm](https://github.com/Kek5chen/devicon/blob/0a06ff51f0a88db2549bd71d63a5c5a2f2dc39d8/icons/pycharm/pycharm-original.svg)
[Rider](https://github.com/Kek5chen/devicon/blob/0a06ff51f0a88db2549bd71d63a5c5a2f2dc39d8/icons/rider/rider-original.svg)
[WebStorm](https://github.com/Kek5chen/devicon/blob/0a06ff51f0a88db2549bd71d63a5c5a2f2dc39d8/icons/webstorm/webstorm-original.svg)

Some icons i only updated, but I only updated the `original` files. The others, except `plain`, seemed outdated as there is no actual `wordmark` for the products officially available anywhere.

Would you like me to create a PR for maintainers to modify to suit the repository guidelines?

Snailedlt:
Hey, create a PR so we can get this one going :)
Jetbrains icons definitely need updating

Kek5chen:
[...]

PR #1291, opened 2022-07-24
update icon: clion (original, plain, plain-wordmark)

kilian-paquier:

- [x] PR name matches the format *update icon: <i>Icon name</i> (<i>versions separated by comma</i>)*. More details [here](https://github.com/devicons/devicon/blob/develop/CONTRIBUTING.md#overview)

**This PR contributes to #1125**

https://www.jetbrains.com/clion/


Panquesito7 (dismissed):
Thanks for contributing! 🚀


__
[sema-logo](https://semasoftware.com/gh) &nbsp;**Summary:** :trophy: This code is awesome&nbsp; | &nbsp;**Tags:** Maintainable

icons/clion/clion-plain-wordmark.svg
```diff
@@ -1,4 +1,3 @@
...
...4 1.123-.844a3.296 3.296 0 0 1 1.72.563l.39-.66a3.993 3.993 0 0 0-2.087-.64zm-18.933.003a2.456 2.456 0 0 0-1.44.434c-.426.288-.75.702-.93 1.187v-1.52h-.869v5.813h.87V80.81c0-1.758 1.028-2.668 2.298-2.668h.07v-.934zm3.32.016a4.84 4.84 0 0 0-2.111.484l.26.727a3.997 3.997 0 0 1 1.761-.426c1.094 0 1.737.535 1.737 1.586v.183a6.037 6.037 0 0 0-1.791-.25c-1.472 0-2.485.653-2.485 1.88v.027c0 1.194 1.09 1.812 2.168 1.804a2.557 2.557 0 0 0 2.108-1v.875h.828v-3.55a2.283 2.283 0 0 0-.594-1.7 2.567 2.567 0 0 0-1.881-.64zm-42.254.078 2.645 5.8c-.348.82-.676 1.09-1.196 1.09a1.853 1.853 0 0 1-.91-.214l-.293.683a2.583 2.583 0 0 0 1.239.293c.874 0 1.437-.438 1.937-1.64l2.436-6.004h-.918l-1.873 4.84-2.11-4.848h-.957zm46.535.008v5.8h.866v-5.8h-.866zm-27.3.609c1.156 0 1.761.918 1.851 1.98h-3.74c.113-1.14.878-1.98 1.889-1.98zm-22.75.043c1.101 0 2.046.875 2.046 2.246v.027c0 1.403-.92 2.239-2.047 2.239a2.18 2.18 0 0 1-2.105-2.254v-.02a2.175 2.175 0 0 1 2.105-2.238zm32.093 1.574h2.487v.012c1.375 0 2.089.496 2.078 1.336v.027c0 .875-.747 1.403-1.934 1.403h-2.63v-2.778zm13.615.64a6.014 6.014 0 0 1 1.708.247v.562c0 .926-.868 1.576-1.958 1.579-.774 0-1.455-.434-1.455-1.157v-.027c0-.731.615-1.203 1.705-1.203z"/>
```
Snailedlt, line 2: Shouldn't this also have the default color as fill?

Snailedlt: Maybe we should have original-wordmark with the default color and plain-wordmark with green color? But then it would be against TOS (I think) since the square in the middle has to be black. Not sure if that applies to us though

I want some more opinions on this though. @Panquesito7 @BenSouchet @kilian-paquier any thoughts on this?

BenSouchet: Hi,
In my opinion the current icons added by @kilian-paquier are good, shapes and colors, my only comment: I would add the `original-wordmark` version with the filled white area on the CL and the oblique line (in the square).

Just like the logo on the clion product homepage:
Screenshot 2022-08-03 at 12 38 25

Snailedlt: Hi @BenSouchet  
Thanks for chiming in!

Okey, but what about this from the [Jetbrains Brand Assets](https://www.jetbrains.com/company/brand/#logos-and-icons-product-logos) page though? 
>The Black Box used in our logos must remain untouched – in terms of both its color and proportions to the text. Make sure that the Black Box always remains black and is never changed (e.g inverted) to another color, even from the official color palette of JetBrains.

Shouldn't we follow it when we can, or at the very least not encourage usage against their guildelines/terms?


This is what I think:
I don't think it's a good idea to fill in the background, nor to use the logo sent above. Mainly because it limits users by choosing the background color for them. Leaving it transparent gives more options. But also because transparent background is one of the requirements when adding icons to devicons as per the [SVG Standards](https://github.com/devicons/devicon/wiki/SVG-Standards). Furthermore I think we should use the "Product Logo" jetbrains provides in their brand guidelines to encourage correct usage of the icons/logos.

BenSouchet: @Snailedlt 
I disagree.
`original` version should always be the exact original logo of the tool / product, for jetbrains logos the white square is filled with white:
Screenshot 2022-08-03 at 15 12 42
Screenshot 2022-08-03 at 15 12 48

Taken from https://www.jetbrains.com/company/brand/

And the screen I share in my previous comment is from the product homepage of clion.

In the quote you shared `Black Box used in our logos must remain untouched` this means with the white part (for the `original` version)

Now for `plain` version these white part must be transparent otherwise in the devicon font generated it will be a full fill square (since for the font colors are ignored)

BenSouchet: And last point with `original` users **cannot** set a new color, `only for `plain` versions

Snailedlt: @BenSouchet Ohh, I see I misunderstood parts of your comment earlier.

To clarify:
When you said ` I would add the original-wordmark version with the filled white area on the CL and the oblique line (in the square).` and showed a picture of the logo, I misread and thought you meant the blue (not the white) background color should be added. Now that I know you meant the white parts inside the black box, I 100% agree with you. Didn't even notice it was transparent, so good catch on your part. 
I thought your main point was to add the blue colored background, or more specifically, that excact logo, which I disagreed with.

So if I understand you correctly now, you want the following change:
1. Add a new `original-wordmark` with the black box untouched (white spots must go back in)
So you mean like this then, right?


I agree with these changes ✔️ 

That said, isn't it still weird that the two plain icons have different colors? Shouldn't both have either black (like jetbrains wants), or green (to be in line with the main color of the logo)?

BenSouchet: > So if I understand you correctly now, you want the following change:
> Add a new original-wordmark with the black box untouched (white spots must go back in)
> So you mean like this then, right?

Yeah exactly 😊

> That said, isn't it still weird that the two plain icons have different colors? Shouldn't both have either black (like jetbrains wants), or green (to be in line with the main color of the logo)?

Since users can edit / change the color of the `plain` versions, it's not that important (in my opinion). I would say to continue to do as it has already been done for the others JetBrains tools / softwares.

I think I saw that @kilian-paquier  chose a unique color for every softwares, and apply this color on the `plain` version, and set black for the `plain-wordmark` version.
It doesn't bother me, on the wordmark version there is more black area than colored area (due to the wording).

Snailedlt: Okey, I've no objections to doing it this way, just thought it looked a bit weird at first glance.

Will wait for @Panquesito to comment here before continuing though :)

kilian-paquier: So, what do we do ? 😄

Snailedlt: I say keep it as it is, unless @Panquesito7 is stronly against that :)

BenSouchet: @kilian-paquier @Snailedlt 
I still think that we should add an `clion-original-wordmark.svg` version with the filled white areas on the square. Like the one you shared in your previous comment @Snailedlt .

Snailedlt: Agreed!
@kilian-paquier Could you add a original-wordmark logo like explained above?

Snailedlt (changes requested):
`clion-original-wordmark.svg` version with filled white areas on the square should be added

kilian-paquier:
> `clion-original-wordmark.svg` version with filled white areas on the square should be added

Ok will do

Snailedlt:
@kilian-paquier while you're at it. Can you do the same with your other PR's? I noticed some of the others were also missing original-wordmark with white inside the square

kilian-paquier:
> @kilian-paquier while you're at it. Can you do the same with your other PR's? I noticed some of the others were also missing original-wordmark with white inside the square

Yep. Also, we merged some JetBrains products already 🙁

Snailedlt:
> > @kilian-paquier while you're at it. Can you do the same with your other PR's? I noticed some of the others were also missing original-wordmark with white inside the square
> 
> Yep. Also, we merged some JetBrains products already slightly_frowning_face

Yeah, I noticed.

Maybe you could make a PR to fix the things we've mentioned in these newer PR's for the already merged icons?

icons/clion/clion-plain-wordmark.svg
Snailedlt: @amacado agreed that this is a good solution, so I think we can go ahead and merge this once `clion-original-wordmark` has been added as specified above.

@kilian-paquier ^

Snailedlt (changes requested):
Only missing a `clion-original-wordmark` version with the official unedited wordmark logo (white parts must be intact), and we're good to go :)

devicon.json
```diff
@@ -1095,20 +1095,23 @@
...
+            "c++",
+            "cpp",
+            "cplusplus"
         ],
         "versions": {
             "svg": [
                 "original",
-                "plain",
-                "plain-wordmark"
+                "original-wordmark"
```
Snailedlt, line 1107: I don't see a reason why these should be removed
```suggestion
                "original-wordmark",
                "plain",
                "plain-wordmark"
```

BenSouchet: @Snailedlt It's normal, @kilian-paquier is correct to remove `plain` from the SVG list, `plain` exists normally only for the font, so `plain` SVG need to be referenced only on the `"font": [` list.

BenSouchet: To nuance my previous message its seems that sometime `plain` versions are also listed in the `"svg"` list, like for example `heroku` or `haskell`.

BenSouchet: @Snailedlt I removed my previous comments , i was wrong and yes your edit request is valid, my bad 🙂

Snailedlt: @BenSouchet I didn't see it, and no worries :)

Snailedlt (approved):
Thank you for your contribution! :rocket:

PR #1367, opened 2022-08-30
docs: improve contributing deprecation notice

Panquesito7:
Things added/changed:

- Improve contributing deprecation notice.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,6 +1,6 @@
 <h1>Contributing to Devicon</h1>
 
-<h2>This guide has been converted to a <a href="https://github.com/devicons/devicon/wiki">Wiki</a>! This document will eventually be removed, please use the Wiki instead.</h2>
+<h2>This guide has been moved to the <a href="https://github.com/devicons/devicon/wiki">Wiki</a> and won't be updated any longer.</h2>
 <br>
 <p>
 First of all, thanks for taking the time to contribute! This project can only grow and live by your countless contributions. To keep this project maintainable, we have developed some guidelines for our contributors.


CONTRIBUTING.md
```diff
@@ -503,4 +503,4 @@ We are running a Discord server. You can go here to talk, discuss, and more with
 | Visual Studio Code | https://code.visualstudio.com/                     | A code editor for editing code                     |
 | vscode.dev         | https://vscode.dev/                                | Visual Studio Code in the browser                  |
 | Iloveimg           | https://www.iloveimg.com/resize-image/resize-svg   | Resizing SVG's                                     |
-| svgviewer.dev      | https://www.svgviewer.dev/                         | View, save, and optimize SVGs                      |
\ No newline at end of file
+| svgviewer.dev      | https://www.svgviewer.dev/                         | View, save, and optimize SVGs                      |
```
Snailedlt, line 506: I don't see a change here. Could you point out what exactly changed on this line?

Snailedlt (changes requested):
Looks good, but for some reason there seems to be an empty change at line 506. Any idea what that is?

CONTRIBUTING.md
Panquesito7: I'm not sure about that, but I didn't make any changes there.

BenSouchet: it's maybe the line ending / line break, check https://stackoverflow.com/questions/1552749/difference-between-cr-lf-lf-and-cr-line-break-types

Snailedlt: I'll approve once this diff is fixed. Invisible changes always leave me with a bad taste in the mouth

Snailedlt (changes requested):

Panquesito7:
> I'll approve once this diff is fixed

I'm not sure that can be fixed. I could try but that's something that I didn't change.

Panquesito7:
I found the issue: it seems there was a missing endline in the original file, and on this file, it was automatically added.
That's OK as some IDEs could have issues without an endline. There should be no issues with that. 🙂

Snailedlt (approved):

PR #1410, opened 2022-10-02
Bugfix/fix in develop labeler

### LINKED ISSUE #1099: [BUG] The `in-develop` labeler is not working due to restricted `GITHUB_TOKEN` access

- [X] Confirm

### Bug description

Currently, the in-develop issue labeler is not working properly (see [this](https://github.com/devicons/devicon/actions/runs/2032211622). The issue is due to the `GITHUB_TOKEN` being limited in scope when used in forked repos. 
image

Since all PRs are from forked repos, it renders the workflow unable to edit the Issues in our repo.

I was going to try [this solution](https://github.com/actions/first-interaction/issues/10#issuecomment-1041402989). It involves giving the workflow access to these permissions to add the label. However, this is not recommended since our workflow involves checking out the repo => might open ourselves up to [pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/). It also seems like we can't override the access for forked repo as seen in this [link](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#modifying-the-permissions-for-the-github_token)

New possible solution: follow the pattern that's being used by our `peek-bot`:
1. The `pull_request` PR saves the PR number/issue number into a file
2. Upload it as an artifact
3. Create a `workflow-run` workflow that runs after the first PR
4. Get the artifact and label our issue in here where we get full GITHUB_TOKEN access. 

Since `workflow_run` workflows has to be in the default branch to work, this would prevent security issues.

### Additional information

For now, we can label or the issues manually after merging. It's a bit annoying but not too difficult.

Snailedlt:

## Features

Fix in-develop labeler by following the steps provided by @Thomas-Boi in #1099
> New possible solution: follow the pattern that's being used by our peek-bot:
> 1. The pull_request PR saves the PR number/issue number into a file
> 2. Upload it as an artifact
> 3. Create a workflow-run workflow that runs after the first PR
> 4. Get the artifact and label our issue in here where we get full GITHUB_TOKEN access.

**This PR closes #1099**

## Notes

This needs to be merged to master in order to work, since it uses `workflow_run` which runs a github actions workflow from the master branch, and the master branch does not yet contain that workflow. Therefore I suggest merging this to develop just before the next release.

The bugfix has been tested on my forks by adding it to master, and then creating a pr from a different fork. The PR can be found [here](https://github.com/Snailedlt/devicon/pull/46), with the accompanying actions run [here](https://github.com/Snailedlt/devicon/actions/runs/3169005151/jobs/5160612079), which successfully labeled the following issue: https://github.com/Snailedlt/devicon/issues/17
image of workflow run

**NB!**
This PR is missing one commit compared to the one I had in my fork. In order to test I had to change the base_url from devicons/devicon to my fork snailedlt/devicon, after the test I removed that temp commit via an interactive rebase and force pushed.


Thomas-Boi (commented):
Hi @Snailedlt,

I looked through your code and everything looks good. I double checked your testing PR/Issue as well and I would like to remark:
1. While PR and Issues are treated the same by GitHub, it might be a good idea to test the whole thing as a PR. Sometime, GitHub Actions differentiate them in unintentional ways. I'd recommend:
  - Make a PR into `develop`  of the forked repo
  - See if the PR gets labelled as `in-develop`
2. Beside testing the above, also test it using a forked repo of the forked repo. This will emulate the scenario that a contributor would be in when they open a PR. What this means in practice:
  - You currently have a forked repo of `devicon`. Call it `devicon-A`.
  - Now fork `devicon-A` to create another forked repo. This is now `devicon-B`.
  - Make a PR from `devicon-B` into `devicon-A`. This will create the cross repo environment that causes the original script to fail in the first place. Thus, if your script works, it should label the PR from `devicon-B` once it's merged into `devicon-A`.

Let me know if the above instructions are too complicated and I can try to explain it better. Other than that, the code looks good and I'm happy to see that someone finally get around to fixing this bug 👍

Snailedlt:
@Thomas-Boi Thanks for butting in with the review, much appreciated 🙇 
1. I tested this, and it worked just fine, but since I was working on other PR's at the same time, I reset the develop branch after that, so that I could base the new branches on it. That's why it looks like it wasn't tested.

2. If I understand your explanation correctly, this is exactly what I did. I might have done a bad job explaining it so I'll try again. 
    1. I made a feature branch on a fork of my fork (I'll call it org-fork from now on). 
    2. Then I made a PR from that org-fork's feature branch into my fork's develop branch. 
    3. When that PR was merged, the `in_develop_labeler_preflight.yml` script ran. Which triggered the `in_develop_labeler.yml` script.
    4. Both ran successfully (see screenshot in this PR's body), and the issue got labeled (highlighted in the screenshot)

I do see that I forgot to rename the script from `in_develop_labeler_preflight.yml` to `on-develop-pr-merge` though, so I guess I'll rename it to that in a new commit. Unless you (or anyone else) has any suggestions for a better name? Naming is hard

Thomas-Boi:
@Snailedlt sweet! Thanks for the explanation. I misunderstood the test PR and thought that it's a PR in the same branch. I'm glad that you already test the cross-repo situation 👍.

As for the naming scheme, I think the current name `pre-flight` is better than `on-develop-pr-merge`. The second one is not clear that it's specifically for `in-develop-labeler`. Just a thought: for the other workflows (`peek` and `build`) I have a `post` in the workflows that run afterwards. Since this one run before, maybe just use a `pre-in-developer-labeler`? It's up to you though. I've considered other naming schemes (`pt1` and `pt2` for each stage, attaching `post` to workflows that run after etc.) but none ever seem like the best answer.

Great work SnailedIt 😄

Snailedlt:
@Thomas-Boi 
It happens :P I would probably have thought the same thing tbh.

As for the naming convention I actually prefer `on-develop-pr-merge` for the exact same reason you don't. That is, that it explains better what it does. Since it's not explicitly for the `in-develop-labeler` in theory, but just in execution. The only thing it does is upload the pr number as an artifact, which I think will be useful for future jobs/workflows too. One example is for example if we wanted to build and test after a merge. Then we could use the same script as a preflight without renaming it. I guess I could call it something else though, maybe `on-develop-pr-merged` or `on-successful-merge-to-develop`

.github/scripts/in_develop_labeler.py
```diff
@@ -4,8 +4,12 @@
 def main():
 	args = arg_getters.get_in_develop_labeler_args()
 	try:
+		#get pr body
+		pr_body = api_handler.get_pr_by_number(args.token, args.pr_num)["body"]
```
Panquesito7, line 8: Shouldn't this be `pr_num`?

.github/scripts/in_develop_labeler.py
```diff
@@ -4,8 +4,12 @@
...
+
 		# find the issue closing line
-		issue_line = [line for line in args.body.split("\n") if line.startswith("**This PR closes")][0]
+		print(pr_body.split("\n"))
+		issue_line = [line for line in pr_body.split("\n") if line.startswith("**This PR closes")][0]
```
Panquesito7, line 12: Users could use multiple close statements. The complete list can be seen [**here**](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword).
To make the process easier, I thought of making an array including all the keywords, then checking if the PR contains one of those keywords.

Snailedlt: Yeah, agreed... though that is for another PR. I'll make an issue for it :D

Panquesito7: > Yeah, agreed... though that is for another PR. I'll make an issue for it :D

I'm not experienced in Python, but I could somehow try doing something if you'd like.

Snailedlt: I think we can improve it past just being keyword based. I made an issue here where we can discuss it further :)
https://github.com/devicons/devicon/issues/1578

I'll resolve this comment now, since I think it should be improved in a separate PR.

.github/scripts/in_develop_labeler.py
Snailedlt: hmm, no I don't think so. 
It gets the `pr_body` using the `get_pr_by_number()` function from the `pr_num` argument (`args.pr_num`), and then choosing the body (`["body"]`)

the pr_body is later used to check for checking if any issues are linked to the PR

Panquesito7 (approved):
Awesome! Thanks. 🚀

.github/workflows/in_develop_labeler_preflight.yml
```diff
@@ -0,0 +1,34 @@
...
+      - name: Setup Python v3.8
+        uses: actions/setup-python@v2
+        with: 
+          python-version: 3.8
+
+      - name: Install dependencies 
+        run: |
+          python -m pip install --upgrade pip
+          pip install -r ./.github/scripts/requirements.txt
+
```
lunatic-fox, line 23: Maybe we can delete this part.
```suggetion
```

Snailedlt: Probably, will have to test it first though

Panquesito7: Not sure that will work. It is needed for the `icomoon_build.py` script, AFAIK.

Snailedlt: @Panquesito7 that sceipt isn't ran from this workflow though. At least as far as I can tell

Panquesito7: Whoops, right. Got confused with another script. 😅
We can remove that and it should work. You can still test if you want, though.

Panquesito7 (dismissed):
👍 once the reviews above are addressed.

.github/workflows/in_develop_labeler_preflight.yml
Snailedlt: Ehh what the heck. Let's just remove this part of the code. It's probably fine, and it doesn't look like I'll have time to test it anytime soon :/

I'll remove the code for python for now and keep the PR open for a bit. Feel free to test it if you want and have time :)

Panquesito7 (approved):
LGTM. Thanks! 🚀

Panquesito7:
Will be merging this for now. If there are any more objections, feel free to create another PR. 🙂

PR #1460, opened 2022-10-15
Check if PR base is develop

### LINKED ISSUE #1458: [FEATURE REQUEST] Automatically check if a PR is being committed into the `develop` branch

- [X] Confirm

### Problem

A few times PR's are being merged into `master` instead of `develop`. If the reviewer doesn't notice this, and merges it directly into master it can cause issues, and at the very least it will be a big timesink to fix it. Ideally we'd want this to never happen.

### Possible Solution

Add a check to the `check-bot` where we check if the PR is going to develop. If it's going to any other branch, it should post a comment telling the PR owner to change the branch to `develop`

### Additional information

_No response_

Snailedlt:

## Features

Check if the base branch is `develop`, and post a PR message if it's not.
image

**This PR closes #1458**

## Notes

~~**Needs to be committed into master before working. Committing it to develop might leave it in a non-working state.**
Therefore I recommend waiting until the next release, and merging this in just before the next release.~~
This is no longer the case, since we didn't have to make any breaking changes to the script after all, thanks to @lunatic-fox's brilliant implementation 💯

### DIFF
--- a/.github/scripts/build_assets/api_handler.py
+++ b/.github/scripts/build_assets/api_handler.py
@@ -65,6 +65,34 @@ def get_merged_pull_reqs(token, page, log_output: FileIO=sys.stdout):
         if merged_pull_req["merged_at"] is not None]
 
 
+def get_pull_req(token, pr_number):
+    """
+    Get a PR based on the PR number
+    See https://docs.github.com/en/rest/pulls/pulls#get-a-pull-request for more details on the parameters.
+    :param token, a GitHub API token.
+    :param pr_number, the number of the pull request.
+    """
+    url = f"{base_url}pulls/{pr_number}"
+    headers = {
+        "Authorization": f"token {token}"
+    }
+    response = requests.get(url, headers=headers)
+    if not response:
+        print(f"Can't query the GitHub API. Status code is {response.status_code}. Message is {response.text}")
+        sys.exit(1)
+    return response.json()
+
+
+
+def get_pr_base_branch(pull_req_data):
+    """
+    Check whether the PR's base is develop. Meaning, if the PR is being committed to develop
+    :param pull_req_data - the data on a specific pull request from GitHub.
+    :return the base ref of the pull_req_data
+    """
+    return pull_req_data["base"]["ref"]
+
+
 def is_feature_icon(pull_req_data):
     """
     Check whether the pullData is a feature:icon PR.
--- a/.github/scripts/build_assets/api_handler.py
+++ b/.github/scripts/build_assets/api_handler.py
@@ -65,34 +65,6 @@ def get_merged_pull_reqs(token, page, log_output: FileIO=sys.stdout):
         if merged_pull_req["merged_at"] is not None]
 
 
-def get_pull_req(token, pr_number):
-    """
-    Get a PR based on the PR number
-    See https://docs.github.com/en/rest/pulls/pulls#get-a-pull-request for more details on the parameters.
-    :param token, a GitHub API token.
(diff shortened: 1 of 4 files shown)


Thomas-Boi (approved):
The PR looks good to me 👍  Code is clean and easy to read. Thank you for working on this feature, this used to happen a lot back then 😆

Snailedlt:
Some PR checks are being skipped, will double check if it's something that needs to be fixed or not.
Converting the PR to a draft while I test things

lunatic-fox:
Hello there!
Perhaps we can use this step to check.
```yml
- name: Message if PR base is invalid
  if: ${{ github.base_ref != 'develop' }}
  run: |
    echo -e "Too bad!\nThis is the wrong PR base!"
    exit 1
```

This step breaks the workflow if the condition is not satisfied.
if-failure

Otherwise the step will be skipped and the others will occur.
if-success

#### The action that I'm using in this test
```yml
name: Check if PR base is develop
on: pull_request
jobs:
  check-pr-base:
    name: Check if PR base is develop
    runs-on: ubuntu-latest
    steps:
      # This step is for test purpose only!
      - name: Echo PR base
        run: |
          echo "PR base is: ${{ github.base_ref }}"

      - name: Message if PR base is invalid
        if: ${{ github.base_ref != 'develop' }}
        run: |
          echo -e "Too bad!\nThis is the wrong PR base!"
          exit 1

      # This step is for test purpose only!
      - name: Check if PR base is valid
        run: echo -e "Great!\nThis is the right PR base!"
```

Snailedlt:
@lunatic-fox hmm, yeah that sounds like a better idea!
Feel free to change my PR accordingly :)

lunatic-fox:
I saw the big picture of this workflow and I think there are issues to my idea.

If this step is `true`, in other words, `develop` is not the base branch, it'll return code 1, which will block the rest of the steps, complete the workflow and begin the other one.

Since this workflow is triggered by a pull request, I'm afraid that it will run just once.

---

It's possible to make this step return a message to `./err_messages/err_messages.txt`, but if the PR has more errors than that one they will not append to this file, because code 1.

Your idea doesn't break the workflow, which is nice, even talking about a major error, because at the end we will have the full report of what is going on with the PR.

---

Tell me what you think. 😀

Snailedlt:
@lunatic-fox no your idea is fine since `on: pull_request` triggers on any PR event, like new commits, labeling the PR, etc. So your workflow will still re-run on any changes to the PR. Besides, if we want to get the other failure reports first we can put the base branch check at the end of the workflow :)


You can read more about on: pull-request [here](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request).

lunatic-fox:
### Input
- Added the step named `Check if PR is develop`, that if is `true`:
  - Records the error message to `./err_messages.txt`
  - Creates the environment variable `wrong_branch`

- Added this expression below to the next 3 steps after, then if the pull request has a wrong base it will skip the icons check.
```yml
if: ${{ !env.wrong_branch }}
```
**Preview**
process

**Comment**
result

---

### Warning
As far as I see, we can't just simply use the `Edit` command to change the branch, since this will not be recognized as an event to rerun the workflow. So, in order to make it work, we have to change the branch and do some `pull_request` event of the list.

**Example**
warning

Snailedlt:
@lunatic-fox Alright, tested and it looks fine now :)
https://github.com/Snailedlt/devicon/pull/65
image

lunatic-fox (approved):
Approved! ✔
A great feature on the way! 🚀

PR #1492, opened 2022-10-31
Run check bot on labeling PR with `bot:check`

Snailedlt:

## Features

**This PR closes NONE**

Didn't create an issue for this since it was quicker to just implement it

## Notes

You can now run the `check-bot` by adding the `bot:check` label to a PR. This is powerful, since it lets us get new PR info (like the title) as input to `check-bot` without creating a new PR. So instead of creating a new PR everytime we fix something, we can instead re-add the `bot:check` label to re-run it with new parameters :)

PS: https://github.com/devicons/devicon/wiki/Our-automated-tasks-and-bots#check-bot needs to be updated with info on how to run the check bot after this PR is merged into master


lunatic-fox (dismissed):
Loved it! 🤩
To me it's approved! ✔

Snailedlt:
@lunatic-fox this is actually a PR I forgot to edit 😅
I wanna change it to a different label, since we use this label to label any icon PR, so the workflow would run twice on each PR. Instead I wanna use a new label (maybe `bot:check`?) to run the workflow. That way we can more easily choose when to run the PR without having to remove the feature:icon label 😁

lunatic-fox:
Hmm... I think the workflow needs to trigger in two events:
- When the pull request is made.
- When a label is created.

Then if the "created" label is, for instance, `bot:check` it'll run the proper steps. Otherwise it should skip the job.

Snailedlt:
@lunatic-fox I changed the label... ready for review again now :)

> Hmm... I think the workflow needs to trigger in two events:
> 
> * When the pull request is made.
> * When a label is created.
> 
> Then if the "created" label is, for instance, `bot:check` it'll run the proper steps. Otherwise it should skip the job.

It already automatically runs on new PR's, provided that the PR title starts with `new icon` or `update icon`, so this PR only adds the additional trigger when the `bot:check` label is added to the PR :)

Panquesito7 (dismissed):
Looks good 🚀 apart from what @lunatic-fox mentioned.

Snailedlt:
@lunatic-fox I didn't understand that explanation fully, but if it's an easy fix, feel free to fix it. If you want me to fix it I'm afraid I'll need to be spoon fed the explanation again :S

Panquesito7:
Will be merging for now. If there are pending things to change, feel free to do that in another PR. 🙂

Panquesito7:
Ping.

Snailedlt:
@lunatic-fox You mentioned an easy fix [here](https://github.com/devicons/devicon/pull/1492#issuecomment-1346820644), but I'm not sure what you mean the issue is. Do you mind fixing it?

Snailedlt:
or should I just merge it as-is?

canaleal:
Has not been updated in a while, closing, but it can be reopened.

Issue #1306, opened 2022-07-26
[OTHER] Extending wiki with tutorials for each recommended tool


Snailedlt:
### I have searched through the issues and didn't find my problem.

- [X] Confirm

### What would you like to share?

It would be nice if we had a wiki page with tutorials for all the recommended tools. This can of course be added incrementally.

@BenSouchet Has already made a quick tutorial for a new tool recommendation here -> https://github.com/devicons/devicon/issues/1279#issuecomment-1195433256

We could follow the same format for the other tools that we have in the [Recommended resources and tools](https://github.com/devicons/devicon/wiki/Recommended-resources-and-tools) wiki page, then add a link back in the table on that page.

#### Example

| Tool Name          | Link                                               | Description & Usage                                |Tutorials|
| :----------------- | :------------------------------------------------- | :------------------------------------------------- |:----|
| Inkscape           | https://inkscape.org/                              | Free desktop application for editing and making SVGs   |[how to foo](https://example.com/), [how to bar](https://example.com/)|
| Visual Studio Code | https://code.visualstudio.com/                     | A code editor for editing code                     |[how to vscode](https://example.com/)|
| vscode.dev         | https://vscode.dev/                                | Visual Studio Code in the browser                  |[how to vscode](https://example.com/)|
| Iloveimg           | https://www.iloveimg.com/resize-image/resize-svg   | Resizing SVGs (this doesn't work with SVGs containing complex elements like gradients, etc.). **Still have to add the attribute `viewBox="0 0 128 128"` manually to your resized files.**|[cool tutorial](https://example.com/)|
| svgviewer.dev      | https://www.svgviewer.dev/                         | View, save, and optimize SVGs                      |[tutorial1](https://example.com/)|
| Vecta.io           | https://vecta.io/nano | Compress and optimize SVGs|[tutorial1](https://example.com/)

#### I propose the following additions

- 1 new wiki page for `tools usage and tutorials`
- A new column in the table on the [Recommended resources and tools](https://github.com/devicons/devicon/wiki/Recommended-resources-and-tools) wiki page

Snailedlt:
@weh That png to svg tutorial you showed in #1428 is relevant here

Snailedlt:
How to unite paths in Inkscape:
1. Ungroup the objects Ctrl Shift G
1. Select all Ctrl A
1. Path > Union Ctrl + +

Snailedlt:
[...]

PR #1527, opened 2022-11-20
update icon: apache (original, plain, line, original-wordmark, plain-wordmark, line-wordmark)

### LINKED ISSUE #1525: [BUG] apache logos are wrong

- [X] Confirm

### Bug description

Official apache logo: https://www.apache.org/logos/res/httpd/httpd_highres.png

Logo in devicon: https://github.com/devicons/devicon/blob/master/icons/apache/apache-original-wordmark.svg

Official logo is vertical and does not have outline. Also font used in text is wrong.

Though it could be just old vs new version issue. But it does need fixing anyway because current logo contains raster images.

Redo logo

### Additional information

This is a split from second post of https://github.com/devicons/devicon/issues/1517

cyberalien:

## Features
Rewrite of Apache logos

**This PR closes #1525**
https://github.com/devicons/devicon/issues/1525

## Notes
Full redesign of logos, based on [official Apache logo](https://www.apache.org/logos/res/httpd/httpd_highres.png)


devicon.json
```diff
@@ -355,7 +355,7 @@
                 "line-wordmark"
             ]
         },
-        "color": "#303284",
+        "color": "#ab0001",
```
Snailedlt, line 358: Same color as the main fill in the `original-wordmark` SVG
```suggestion
        "color": "#aa0001",
```

Snailedlt (changes requested):
Those icons are looking nice!

In addition to changing the color, please remove the width and height attributes, as well as optimizing the icons with SVGOMG by following [this](https://github.com/devicons/devicon/issues/1306#issuecomment-1296015861) guide.

We should add checks for this, and automate the optimization with SVGO (I'll make issues for doing that)

devicon.json
cyberalien: Sure. Done

Snailedlt:
Looks like you removed the viewbox instead of the height and width attributes. Whoopsie 🙊 

It can probbaly be fixed by just replacing
```svg
width="128" height="128"
```
With
```svg
viewbox="0 0 128 128"
```

cyberalien:
Doh!

Fixed :)

Snailedlt:
You're so fast! :100:

Snailedlt:
Te line-wordmark and plain-wordmark versions have these small defects :/

Screenshot_20221121_001227

Do you know how to fix them? If not, you can tag lunatic-fox to get help fixing it, since he knows how to do it :)

cyberalien:
I've checked shapes, those bugs don't exist. So looks like icomoon bug. My guess is it can't handle compressed paths that SVGO generates, which is a common problem with old parsers and often it looks just like that.

Updated icons with cleaned up path, hoping it fixes issue.

cyberalien:
Hm. Bad guess. Something else is bugged :(

Snailedlt:
Yeah, I think those lines are usually the result of merging paths, causing paths to overlap. @lunatic-fox knows more about it, and can probably fix it when he has time.

@lunatic-fox maybe you could also make a little guide on how we can fix these issues?

cyberalien:
Found a way to debug it on icomoon by creating custom font and importing icons. Icon does show that error there, but only after SVGO optimisation. Icon exported directly from Figma does not have that issue.

cyberalien:
Should be fixed now

Snailedlt:
Ahh, yes the issue seems to be the combining of the paths. Sadly, line and plain icons only containing one path is a requirement from our side though. Luckily this is usually fixable by offsetting the paths a bit. I haven't found a good way to fix it consistently, but I think @lunatic-fox has, so hopefully he can share some knowledge about this

cyberalien:
I've combined paths, it works fine. Issue is somewhere else.

In code I've used to clean up icons, I did run SVGO, but with the following plugins disabled: `removeHiddenElems`, `convertShapeToPath`, `convertEllipseToCircle`, `convertPathData`, `mergePaths`, `reusePaths`. Its a preset I'm using for animated icons, which is designed to not mess with shapes because SVGO is very bad at working with SVG 2 animations, messing up icons in 99% of cases.

So one of them causes issue. Biggest suspect is `convertPathData`.

Snailedlt:
I see, maybe [SVGOMG](https://jakearchibald.github.io/svgomg/) has the same options then. Worth checking out and updating the guide I linked to earlier if it does :)

icons/apache/apache-line-wordmark.svg
```diff
@@ -1 +1 @@
...
...45.6386 65.7632 45.6486 65.804C45.6652 65.8725 45.6582 65.8797 45.5589 65.897C45.4209 65.9208 38.5957 65.9533 38.5957 65.9301C38.5957 65.9205 38.8285 65.351 39.1131 64.6646C39.8838 62.8054 40.5387 61.2243 41.0029 60.1017C41.2307 59.5509 41.544 58.7956 41.6992 58.4232C41.8544 58.0508 42.0124 57.6636 42.0505 57.5628C42.1611 57.2697 42.2228 57.2327 42.2744 57.4288ZM72.1366 57.5557C72.2337 57.7494 72.4378 58.2465 72.8287 59.2413C73.0177 59.7223 73.1982 60.1793 73.2299 60.2569C73.2615 60.3345 73.3436 60.5376 73.4123 60.7083C73.7366 61.5137 74.0944 62.3833 74.2597 62.7677C74.3597 63.0004 74.4702 63.2733 74.5051 63.3742C74.54 63.4751 74.6142 63.6655 74.6699 63.7974C74.7255 63.9292 74.8467 64.2339 74.9391 64.4744C75.0315 64.7149 75.1976 65.1319 75.3081 65.401C75.4187 65.6701 75.5011 65.8983 75.4914 65.9081C75.4648 65.9348 70.0973 65.9513 69.2088 65.9274C68.7852 65.9161 68.4385 65.8961 68.4385 65.883C68.4385 65.86 68.7009 65.2261 69.9734 62.1752C70.31 61.3684 70.6268 60.6067 70.6775 60.4826C70.8487 60.0638 71.2179 59.1888 71.4929 58.5501C71.6433 58.201 71.7982 57.8265 71.8372 57.7179C71.9328 57.4517 71.9508 57.4217 72.0145 57.4217C72.0518 57.4217 72.091 57.4648 72.1366 57.5557Z"/></g></svg>
```
Snailedlt, line 1: this still has multiple `path` elements, but I guess since they are grouped, it can be considered united?

icons/apache/apache-plain-wordmark.svg
```diff
@@ -1 +1 @@
...
...45.6386 65.7632 45.6486 65.804C45.6652 65.8725 45.6582 65.8797 45.5589 65.897C45.4209 65.9209 38.5957 65.9533 38.5957 65.9301C38.5957 65.9205 38.8285 65.351 39.1131 64.6646C39.8838 62.8054 40.5387 61.2243 41.0029 60.1017C41.2307 59.5509 41.544 58.7956 41.6992 58.4232C41.8544 58.0508 42.0124 57.6636 42.0505 57.5628C42.1611 57.2697 42.2228 57.2327 42.2744 57.4288ZM72.1366 57.5557C72.2337 57.7494 72.4378 58.2465 72.8287 59.2413C73.0177 59.7223 73.1982 60.1793 73.2299 60.2569C73.2615 60.3345 73.3436 60.5376 73.4123 60.7083C73.7366 61.5137 74.0944 62.3833 74.2597 62.7677C74.3597 63.0004 74.4702 63.2733 74.5051 63.3742C74.54 63.4751 74.6142 63.6655 74.6699 63.7974C74.7255 63.9292 74.8467 64.2339 74.9391 64.4744C75.0315 64.7149 75.1976 65.1319 75.3081 65.401C75.4187 65.6701 75.5011 65.8983 75.4914 65.9081C75.4648 65.9348 70.0973 65.9513 69.2088 65.9274C68.7851 65.9161 68.4385 65.8961 68.4385 65.883C68.4385 65.86 68.7009 65.2261 69.9734 62.1752C70.31 61.3684 70.6268 60.6067 70.6775 60.4826C70.8487 60.0638 71.2179 59.1888 71.4929 58.5501C71.6433 58.201 71.7982 57.8265 71.8372 57.7179C71.9328 57.4517 71.9508 57.4217 72.0145 57.4217C72.0518 57.4217 72.091 57.4648 72.1366 57.5557Z"/></g></svg>
```
Snailedlt, line 1: Same here

icons/apache/apache-line-wordmark.svg
cyberalien: Added another commit that merges them in one path, it seems to work correctly on icomoon.

lunatic-fox:
Hello there!
I searched a little and shouldn't we be basing the icons on this one?



***Source: [APACHE SOFTWARE FOUNDATION](https://www.apache.org/logos/#)***

---
Perhaps, you're doing the base on this other one:



***Source: [APACHE HTTP SERVER](https://www.apache.org/logos/#)***

From [**Wikipedia**](https://en.wikipedia.org/wiki/Apache_HTTP_Server#/media/File:Apache_HTTP_server_logo_(2019-present).svg) we can also see this info about "Apache HTTP Server".
> The Apache Software Foundation - Vectorised by [Vulphere](https://commons.wikimedia.org/wiki/User:Vulphere) based from https://www.apache.org/logos/

### But what's the difference?
"Apache Software Foundation" logo has few nodes compared to "Apache HTTP Server", which reinforce the idea that it's the original one.

***Apache Software Foundation***
apache-software-foundation

***Apache HTTP Server***
apache-http-server

The excess of nodes can be a problem when converted to font. And some segments have unnecessary curve nodes within a line segment.

As far as I see when converted to font that segments can be misinterpreted and resulting in spikes or bugged lines.

Changing the word "APACHE" based on "Apache Software Foundation" logo and optimizing in SVGO, they have no issues as seen below.
line-wordmark-result

plain-wordmark-result

I think you can rebase the icons and you will have no problem. But if that happen for some oddly reason I'll be helping ASAP. 🙂

cyberalien:
I agree. The simpler the shape is, the better and that source should help a lot.

cyberalien:
Redesigned it, removed old commits that are now irrelevant, optimised and tested on icomoon.

icons/apache/apache-line-wordmark.svg
```diff
@@ -1 +1 @@
...
... 56.1365 17.6054 55.9005 17.5045 55.5359C17.4036 55.1714 17.5843 54.7878 17.9297 54.6335C17.9622 54.619 17.9958 54.6042 18.0305 54.5889C18.6089 54.3341 19.4731 53.9535 20.3691 52.6236C21.9939 50.2121 23.1829 47.5169 23.5194 45.7099C23.8378 44.0002 23.5844 42.8108 23.194 42.0426C23.0512 41.7617 22.7854 41.4211 22.462 41.1571C22.1291 40.8854 21.8278 40.7685 21.6143 40.7685C21.4815 40.7685 21.2835 40.8239 20.9635 41.052C20.6417 41.2813 20.2747 41.6276 19.8076 42.0999C18.9002 43.0174 18.3426 43.6939 17.8549 44.3078C16.1807 46.4151 14.1528 49.2142 12.1811 52.3762C9.03342 57.4242 6.06446 63.4163 4.88715 69.3122C4.64613 70.5191 4.47754 71.7115 4.40103 72.8936C4.37999 73.2187 4.15538 73.4945 3.84197 73.5813C4.04413 74.0814 4.21241 74.6138 4.29272 75.1359ZM7.12175 73.3741C7.11465 73.3935 7.10755 73.4128 7.10047 73.4322C6.68369 74.5706 6.30671 75.6272 5.98062 76.5651C7.11007 76.5456 8.13238 76.3616 8.98019 75.9381C8.69331 75.8755 8.4615 75.6511 8.39645 75.3548C8.31943 75.0039 8.49694 74.6468 8.82312 74.4963L8.82858 74.4938C9.28675 74.2824 9.74138 74.0727 10.194 73.7843C9.43257 73.9175 8.59834 73.9159 7.66677 73.7607C7.42862 73.7211 7.23152 73.574 7.12175 73.3741Z" clip-rule="evenodd"/></svg>
```
Snailedlt, line 1: Two path elements again (should be one) ;)

Snailedlt (changes requested):

Snailedlt:
Thanks for the input @lunatic-fox. Just out of curiosity though... how did you highlight the nodes of the icon? Inkscape?

lunatic-fox:
> Thanks for the input @lunatic-fox. Just out of curiosity though... how did you highlight the nodes of the icon? Inkscape?

Yeah, it was on Inkscape. 🙂

icons/apache/apache-line-wordmark.svg
cyberalien: Done

cyberalien: Actually... no. That broke icon. Doh

Snailedlt (dismissed):
Looks good! ✔️ 
Thank you for another great contribution 🚀

lunatic-fox:
I got a little break here from research. 😅

And I think I didn't express myself right.
Well, I was asking if we should use the entirely logo, like, with the words "The Apache Software Foundation TM". I know is far way better the way it is now, because they're good to read even in small proportions, but I not sure about modifying it.

Another option would be preserve the font wordmark versions the way it is and do the `original-wordmark` version as the logo will all the features, since it's used as icon and not as font.

Let me now what you think @cyberalien and @Snailedlt.

I'll be answering back until tomorrow. 😉

cyberalien:
I think that would make logo unreadable, unless scaled to really big icon, but then it would be too high, so unusable.

Snailedlt:
Both ways work, but like cyberalien said, it quickly becomes unreadable on smaller sizes. That's not that big of a deal imo though, since we have the non-wordmark version available too. 

Up to you guys what you decide on using, but I think we should add `The Apache Software Foundation` as an altname while we're at it :)

Snailedlt:
```json
altnames: [
    "The Apache Software Foundation®"
]
```

lunatic-fox:
> ```json
> "altnames": [
>     "The Apache Software Foundation™"
> ]
> ```

For sure! 😅

cyberalien:
Is there a standard for `altnames` field?

I've checked file, almost all entries are alternative icon names in lower case, not project name. Few items are project names in mixed case, few items are identical to name but with spacing. What I haven't found though is any entry with `™` and technically many names there are trademarks.

Snailedlt:
@cyberalien nope, there are no naming standards for it. Basically altnames is used together with the name in the search field on the website. So it's only to make the icon more searchable atm. 

Might use it for something more in the future though

devicon.json
```diff
@@ -335,7 +335,10 @@
     },
     {
         "name": "apache",
-        "altnames": [],
+        "altnames": [
+            "The Apache Software Foundation™",
+            "The Apache Software Foundation"
+        ],
```
lunatic-fox, line 341: Maybe the one with trademark character is unnecessary, isn't it @Snailedlt? 🤔
```suggestion
        "altnames": [
            "The Apache Software Foundation"
        ],
```

Snailedlt: Hmm, yeah we only need one of them.  I doubt anyone would use tm in a search, but it is supported, so either one is fine, but no need for both since it's only used for searching :)

Snailedlt (approved):

devicon.json
```diff
@@ -355,7 +357,7 @@
                 "line-wordmark"
             ]
         },
-        "color": "#303284",
+        "color": "#aa0001",
```
Snailedlt, line 360: Is this the right color?
Seems to me like `#BD202E` is the most dominant color in the wordmark icon

Snailedlt (changes requested):

devicon.json
cyberalien: It is from text, not icon. Icon has bunch of gradients.

Snailedlt: I couldn't find `#aa0001` anywhere in the icon's code. Is it the average of the gradient or something?

cyberalien: Double checked it. Actually, logo use neither of those colors, main color for text varies, depending on which logo is used as base.

In icon used in original pull request it is actually `#aa0000`: https://svn.apache.org/repos/asf/comdev/project-logos/originals/httpd.svg

However, due to issues with "A" letter, I've rebuilt icon using this as source: https://svn.apache.org/repos/asf/comdev/project-logos/originals/foundation.svg and it uses `#cb2533` for text.

So correct color is `#aa0000` or `#cb2533`, depending on which Apache project it refers to.

Snailedlt: Ahh, right.
I suggest using `#cb2533` then

```suggestion
        "color": "`#cb2533`",
```

cyberalien: Sure... but it went a bit wrong. Noticed suggestion, clicked button to commit it, only later noticed that color in suggestion included quotes.

cyberalien: Fixed that in a separate commit, force-pushed to get rid of bad commit.

lunatic-fox: The issue happens when I said [here](https://github.com/devicons/devicon/pull/1527#issuecomment-1321879664) about APACHE HTTP SERVER and the text of this one is filled with `#aa0000`, which is practically the same as `#aa0001`, but APACHE SOFTWARE FOUNDATION, which I based my fix, have `#BD202E` as the major color.
So I'll be re-suggesting using `#BD202E`

cyberalien: "Apache Software Foundation" logo uses `#cb2533` for text.

Color in icon is irrelevant because there is no fixed color. It is all gradients.

lunatic-fox (approved):
I was mixing up the gradient color with text color, sorry guys, my bad. 😅
I also ungroup some versions. Hope that's ok!

Approved! ✔

Snailedlt (approved):

PR #1537, opened 2022-11-20
update icon: kaggle (original-wordmark)

### LINKED ISSUE #1523: [BUG] Bad code in kaggle-original-wordmark

- [X] Confirm

### Bug description

Shape uses this clip-path: `clip-path="url(#kaggle-original-wordmark-b)"`, but it does not exist in icon.

Probably remove clip-path because icon seem to be fine without it

### Additional information

_No response_

lunatic-fox:

## Features
- This PR fixes problems related to Kaggle `original-wordmark` icon

**This PR closes #1523**


Snailedlt (dismissed):
LGTM! :heavy_check_mark:

Snailedlt:
@cyberalien can you also take a look at this PR please? :)

cyberalien:
There is big padding around logo, so I suggest to scale it up to fill an entire area.

Why is there no `plain` version? I understand that it matches `original`, but does font parser know to use `original` or will it skip icon?

Snailedlt:
@cyberalien Yes, the original and olain versions are linked through devicon.json's alias object: https://github.com/devicons/devicon/blob/fc34f1b4de1e2a9ce93e7df72f1e72696073b411/devicon.json#L4200

You can read about it here: https://github.com/devicons/devicon/wiki/Updating-%60devicon.json%60#aliases-and-aliasobj

cyberalien:
Thanks.

This is a screenshot to show what I mean about padding.

Frame is 128x128, but icon inside is 63.12x110.6 and has seemingly random spacing on all sides.

In many icon sets icons do have spacing, but it is strictly defined in documentation and is usually 2px (1px for circles) for 24x24 icons. Some icon sets don't use padding. But whatever it is, it is consistent for all icons in icon set.

When using font users don't really notice it. Icon fonts render blurry are aligned like font, but when using icons as SVGs, it can be noticed. When icon is scaled down to something that fits it, for example 32x32 is a normal icon size for 128x128 icon set, if icon has no padding or 4/8px padding, edges look sharp. This icon won't look sharp no matter how it is scaled.

kaggle

Snailedlt:
@cyberalien good explanation, I get the point now, thanks!
I agree, we should enforce more consistent padding in our icons, what do you personally prefer? 1px, 2px, 0px? I wanna hear your opinion since you obviously have a lot of experience with SVG icon libraries through your work with Iconify and other libraries :)

On a second note: 
It would take quite some time to resize all the previous icons, so I don't see it happening any time soon, but for new icons, it should be possible. It can probably also be made much easier to review with the right tooling :)

cyberalien:
I think no padding would fit better. Then icon would scale correctly regardless of what resolution user has.

Padding of 1 or 2 pixels works great for icon sets designed on small grid for specific resolution. Material Design Icons guidelines have excellent explanation with examples: https://m2.material.io/design/iconography/system-icons.html#grid-and-keyline-shapes

But I think those guidelines won't fit this icon set. Logos aren't pixel perfect, grid is large. So I suggest no padding, scaling icon to use maximum available space.

lunatic-fox:
Changing in next commit! 🙂

cyberalien:
Perfect!

Snailedlt (approved):
LGTM! :heavy_check_mark: 
Thanks for the contribution @lunatic-fox and @cyberalien :100:

Issue #1549, opened 2022-11-20
[FEATURE REQUEST] Automatically check if .svg files contain widrh and height attributes


Snailedlt:
### I have searched through the issues and didn't find my problem.

- [X] Confirm

### Problem

Icons should use the `viewbox` attribute instead of the `width` and `height` attributes. Atm we have to check this manually when reviewing a PR, but it can easily be automated with `check-bot`

### Possible Solution

Add checks to `check-bot` and respond with an error message if any .svg files in a PR contain width and/or height attributes

Issue #1548, opened 2022-11-20
[FEATURE REQUEST] Automatically check if SVGs contain raster images


Snailedlt:
### I have searched through the issues and didn't find my problem.

- [X] Confirm

### Problem

Currently raster images are allowed in the SVG's by the check-bot. This should not be the case, since raster images are not scalable, and therefore don't belong in our icons.

### Possible Solution

Add checks to `check-bot`. The bot should check for <img> and <image> tags

### Additional information

See #1517 for some examples and more info

lunatic-fox:
Hello there! Perhaps, we can use this regular expression.
```regexp
/="data:image\/.+;base64,/gm
```

Well, let me try to convert it to Python: 😅
```python
import re

# A mockup of a SVG.
svg_body = "<svg>...</svg>"

# Keeps the amount of images/raster founded.
is_img = len(re.findall("=\"data:image\/.+;base64,", svg_body))

if is_img > 0:
  err_msg.append(f"- `{svg_path.name}` has an image/raster within it.")
```
> *This is my first time ever coding in that programming language.*

Snailedlt:
@lunatic-fox no need to re-invent the wheel here. I think we can just re-use what SVGO has here: https://github.com/svg/svgo/blob/ae32acf5144be318872a45f49b3ece50c0a4cb18/plugins/removeRasterImages.js#L24

lunatic-fox:
Hmm, the idea came from [here](https://github.com/devicons/devicon/pull/1554#discussion_r1030304788). They're regex is a step before and it will totally work.

The full regular expression with common image extensions would be something like this:
```regexp
/<image.+(xlink:)?href="data:image\/(png|jpg|jpeg|gif|tiff|svg\+xml|bmp);base64,.+"\/>/gm
```

But since there are more extensions we could reduce to this:
```regexp
/<image.+(xlink:)?href="data:image\/.+;base64,.+"\/>/gm
```

Or to simplify everything we can just ask if the svg has the `<image>` tag, that may be more than enough to know if there's a image/raster. So we can reuse [line 22](https://github.com/svg/svgo/blob/ae32acf5144be318872a45f49b3ece50c0a4cb18/plugins/removeRasterImages.js#L22) instead. 🙂

Snailedlt:
Yeah, seems like a good idea to just generally not allow the image element. That way we disallow raster images and other linked images, so we have the entire source code in our repo instead of linking to external images.

Snailedlt:
[...]

PR #1411, opened 2022-10-02
add feature:icon and enhancement labels automatically

Snailedlt:
## Double check these details before you open a PR

- [x] PR does not match another non-stale PR currently opened

## Features

add ![][~feature:icon] and ![][~enhancement] labels to pull_request_template.md, so that the labels are added automatically when users copy the link

**This PR closes #1399**

## Notes

I think this needs to be merged to master before it takes effect

[~feature:icon]: https://img.shields.io/badge/-feature:icon-6cc43c.svg
[~enhancement]: https://img.shields.io/badge/-enhancement-84b6eb.svg

PR #1405, opened 2022-10-02
Feature/sort devicon json and add sort check

Snailedlt:
## Double check these details before you open a PR

- [x] PR does not match another non-stale PR currently opened

## Features

1. Sort and re-indent `devicon.json`
2. Add check to `check_icon_pr.py` to see if devicon.json is still sorted

**This PR closes #1337, #1327**

## Notes

Please check both if the functionality looks sound, and if the error messages can be improved

PR #1554, opened 2022-11-23
add feature: svgo auto optimization to pull requests

### LINKED ISSUE #1547: [FEATURE REQUEST]: Automatically optimize icons with SVGO

- [X] Confirm

### Problem

Every contributor has to optimize the icons they add. The main way to do this is following [this]() guide, to optimize the icons using SVGOMG (a SVGO UI tool). It would be much better if this process was automated

### Possible Solution

Make a workflow to optimize icons using SVGO. If there is no GitHub Action to do this already, we should consider making one.

1.The workflow should use the same options that are recommended in the aforementioned guide. 
1. The workflow should optimize all .svg icons in a automatically on each push to the PR branch.
1. If we choose to make a GitHub action, the action should be parameterized

### Additional information

_No response_

lunatic-fox:

### In `optimize_icons.yml`
This workflow runs only on pull requests using [**SVGO action**](https://github.com/marketplace/actions/svgo-action) and [**git-auto-commit Action**](https://github.com/stefanzweifel/git-auto-commit-action).

### In `svgo.config.js`
Converts `style` to inline attributes and removes dimensions, raster images and possible script elements.

In `removeAttrs` we can add some unnecessary atributes to be removed. Currently it will remove:
- Empty `fill`
- `stroke` and `stroke` derived attributes like `stroke-width`
- Attributes that begin with `data`
- `enable-background`
- Some that we should handle with care
  - `xml:space`
  - `fill-rule`
  - `clip-rule`

**This PR closes #1547, #1549 and partially #1548**


.github/scripts/svgo.config.js
```diff
@@ -0,0 +1,53 @@
...
+        overrides: {
+          convertStyleToAttrs: true,
+        }
+      }
+    },
+    {
+      name: 'removeRasterImages',
+      params: {
+        overrides: {
+          removeRasterImages: true,
```
Snailedlt, line 26: I think we want to just let the user know that rasterImages can't be used instead of removing them. 

At least we shouldn't remove them automatically imo. It makes it harder to review, since we have to check the previeous commit to know if parts of the logo got removed. If say the shadow was a raster image, and it was removed, then the icon has been modified too much to be usable as an `original` version.

For these reasons think it would be better to just add a raster image check to the `check-bot`, so that we can't merge PRs with raster images in the icons

Snailedlt (changes requested):
Looking forward to this feature, but it's not quite ready yet

.github/scripts/svgo.config.js
```diff
@@ -0,0 +1,53 @@
...
+      name: "removeAttrs",
+      params: {
+        attrs: [
+          'path:fill:none',
+          'xml.space',
+          'enable-background',
+          '^data.+',
+          'stroke.*',
+          'fill-rule',
+          'clip-rule'
```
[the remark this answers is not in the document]
lunatic-fox: `removeDimensions` does it, removing `height` and `width`, and add `viewBox` attribute if is missing.

About second questions. My intention was to read some feedback about them. Then we can convert the settings to a more basic state.

Strokes shouldn't be allowed, but we should be able to see the error first, so I'll be doing some adjustments.

I'll preserve just `path:fill:none` from this plugin, since it cause no problem at all.

.github/scripts/svgo.config.js
lunatic-fox: I'm ok with it. Removing this plugin in next commit! 🙂
> *That other bot to check raster images, should be easy to do, but if it's in Python I rather leave it to other person who knows better how to work with that language.* 😀 *Still, I can help with the Regular Expressions.* 😉

.github/scripts/svgo.config.js
```diff
@@ -0,0 +1,53 @@
...
+    {
+      name: 'removeDimensions',
+      params: {
+        overrides: {
+          removeDimensions: true,
+        }
+      }
+    },
+    {
+      name: 'convertStyleToAttrs',
```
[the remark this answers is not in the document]
lunatic-fox: That's a lot of style. 😎😂
I think `inlineStyles` can group everything by itself, so we would only need to remove this plugin. However I'll be testing it first.
Then, if it works like that, I'll be removing this plugin in next commit, otherwise I'll be incorporating your idea to this set. 😁

Snailedlt (approved):
Looks good as a starting point now. If this has been tested after the latest commit I'm okey with merging it in.

Good job @lunatic-fox ✔️

Snailedlt:
@lunatic-fox This needs to be merged to master for the script to work right?

lunatic-fox:
> @lunatic-fox This needs to be merged to master for the script to work right?

Actually no, it will work as soon as implemented like #1405. 😀

For instance, when the contributor makes a PR, the `new branch` will be based on `develop` branch. And this `new branch` going to be pull requested to `develop`. In summary, both branches have the workflow, the only exception would happen if the contributor base the PR on an old version of `develop` branch.

Snailedlt:
But doesn't the workflow get the config file from the default (master) branch? At least that's the default behaviour for executing scripts from a workflow file iirc

.github/scripts/svgo.config.js
```diff
@@ -0,0 +1,31 @@
+module.exports = {
```
Panquesito7, line 1: Shouldn't we provide this on the root directory? I think it'd be better there, IMO.

lunatic-fox:
As far as I tested, this new feature will work when merged to `develop`.

However, I found some issues when testing.
- ✔ f2175f3858f7e9994d5110046c475c188706ab5d - Fix the bot permission to write and make a new commit.
- ❌ We still need to create a short simple script to add a new line on each optimized SVG file, since SVGO does not do that. This step needs to occur after `ericcornelissen/svgo-action@v3` step.
- ⚠ Maybe, in order to be more selective, we could create a script to look if the file has more than 2 lines, if so activate this action, otherwise skip it. 🤔

Panquesito7:
> We still need to create a short simple script to add a new line on each optimized SVG file, since SVGO does not do that. This step needs to occur after `ericcornelissen/svgo-action@v3` step.

See https://github.com/devicons/devicon/issues/1624 for more information about this.

lunatic-fox:
> But doesn't the workflow get the config file from the default (master) branch? At least that's the default behaviour for executing scripts from a workflow file iirc

As far as I got, the workflow files come from the branch that you're making the PR, that's why the #1332 works even when this feature is not implemented on `master` yet, because contributors are making PRs from `develop` that has this workflow file.

The issue #1411 is some exception and it's more related to templates, not to workflows. 😅

.github/scripts/svgo.config.js
lunatic-fox: In my point of view I don't see why, since this script is related to a specific workflow, differently, for instance, of `gulpfile.js` that is related to the whole project. 🤔

Snailedlt: Are we talking about the file location? If so I agree with @lunatic-fox. Scripts only used in the workflow should be stored in the workflow/script folder :)

Snailedlt (approved):

.github/scripts/svgo.config.js
```diff
@@ -0,0 +1,31 @@
...
+    },
+    {
+      name: 'removeScriptElement',
+      params: {
+        overrides: {
+          removeScriptElement: true,
+        }
+      }
+    },
+    {
```
Panquesito7, line 22: Shouldn't we remove strokes as well?

Panquesito7 (approved):
Otherwise, this is looking great. Thanks! 🚀

.github/scripts/svgo.config.js
Snailedlt: No. The reason is that this might change the icon. In such cases we should instead tell the user that the stroke needs to be removed in another github workflow.

Relevant PR comment: https://github.com/devicons/devicon/pull/1554#discussion_r1030321954

PR #1595, opened 2022-12-22
update icon: qt (original)

lunatic-fox:

## Features
Unify paths into one, removing `fill="none"` subtraction method and "stroke*" attributes.

**This PR closes NONE**

## Notes
`fill="none"` subtraction method can lead to a solid square in final font instead of the icon.


Snailedlt (approved):
LGTM! ✔️

PR #1605, opened 2023-01-03
new icon: scikitlearn (original, original-workmark, plain, plain-wordmark, line, line-wordmark)

ConX:
## Double check these details before you open a PR

- [x] PR does not match another non-stale PR currently opened
- [x] PR name matches the format *new icon: <i>Icon name</i> (<i>versions separated by comma</i>)*. More details [here](https://github.com/devicons/devicon/blob/develop/CONTRIBUTING.md#overview)
- [x] PR's base is the `develop` branch.
- [x] Your icons are inside a folder as seen [here](https://github.com/devicons/devicon/blob/develop/CONTRIBUTING.md#organizational-guidelines)
- [x] SVG matches the standards laid out [here](https://github.com/devicons/devicon/blob/develop/CONTRIBUTING.md#svgStandards)
- [x] A new object is added in the `devicon.json` file as seen [here](https://github.com/devicons/devicon/blob/develop/CONTRIBUTING.md#-updating-the-deviconjson-)

**This PR closes #1032**

## Link to prove your SVG is correct and up-to-date.

https://scikit-learn.org/stable/about.html#artwork

PR #1616, opened 2023-01-06
new icon: scikitlearn (original, plain, line)

### LINKED ISSUE #1032: Icon request: Scikit Learn
### About the icon

A complete lib for machine learning

### Links and sources

https://scikit-learn.org/stable/

### Additional information

_No response_

ConX:

**This PR closes #1032**
This was submitted a few days ago(#1605) but was closed to fix the branches.

https://scikit-learn.org/stable/about.html#artwork


devicon.json
```diff
@@ -7814,6 +7814,31 @@
...
+        "name": "scikitlearn",
+        "altnames": [
+            "scikit-learn",
+            "scikit",
+            "sklearn"
+        ],
+        "tags": [
+            "machine-learning",
+            "python"
+        ],
```
lunatic-fox, line 7827:
```suggestion
        "tags": [
            "machine-learning",
            "python",
            "tool"
        ],
```

lunatic-fox (commented):
Hello there!
I think we can also add `tool` to tags.
I've optimized your icons using [SVGOMG](https://jakearchibald.github.io/svgomg/). Hope that's ok! 😁

devicon.json
ConX: Thank you @lunatic-fox, I will start doing the same and optimize them prior to submission! I thought that optimization was part of the GitHub Actions.

lunatic-fox: No problem! And you're right to think that way, but we're still developing that optimization to be a proper GitHub action. 🙂

devicon.json
```diff
@@ -7814,6 +7814,31 @@
...
+    {
+        "name": "scikitlearn",
+        "altnames": [
+            "scikit-learn",
+            "scikit",
+            "sklearn"
+        ],
+        "tags": [
+            "machine-learning",
+            "python"
```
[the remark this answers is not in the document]
ConX: I also think that this will be very small for a square viewbox. Moreover, I am still not clear on which version is which. Although [this page](https://github.com/devicons/devicon/wiki/SVG-Versions) is useful for most cases, there are cases where it's not as clear what to choose.

Initially, I thought that the logic is "if it has letters is a `-wordmark,`" but there are some cases (e.g., AWS, scikitlearn) where there isn't a version without the letters. What would help is to add a flowchart to the page on how to choose the version. 

@Snailedlt Let me know if you need me to rename any files here, and I will do so.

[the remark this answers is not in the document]
Snailedlt: @ConX Agreed, it's not always clear which version to choose. 

In addition to the rules in the [SVG-Versions](https://github.com/devicons/devicon/wiki/SVG-Versions) wiki page the rules for wordmark icons is like this:
1. If the icon contains text, then it should have the `wordmark` suffix.
    - If there is no "non-wordmark" version, and the icon is used in the font, there should be an alias between the `wordmark` and the `non-wordmark` versions like so:
```json
        "aliases": [
          {
            "base": "plain-wordmark",
            "alias": "plain"
          },
        ]
```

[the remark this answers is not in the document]
Snailedlt: Though I'm not sure why we only have aliases for fonts. Maybe we can just add them to all versions to make it easier?

Panquesito7 (dismissed):
LGTM! Thanks. 🚀

Snailedlt (approved):

lunatic-fox (dismissed):
Great! ✔

devicon.json
```diff
@@ -7921,6 +7921,33 @@
...
+                "plain-wordmark",
+                "line-wordmark"
+            ],
+            "font": [
+                "plain-wordmark",
+                "line-wordmark"
+            ]
+        },
+        "color": "#f89939",
+        "aliases": []
```
Panquesito7, line 7949: Now that I think about it: shouldn't we add aliases here?

Snailedlt (changes requested):
See panquesito's comment

devicon.json
Snailedlt: You're right, nice catch!

There are two alternatives 

1. `wordmark` should be removed from from the filename, and aliases should be added for wordmark... So like this:

base: original,
alias: original-wordmark

base: plain
alias: plain-wordmark


... Etc

2. Remove wordmark from the icon names, and add the wordmark versions I suggested in this comment: https://github.com/devicons/devicon/pull/1616#pullrequestreview-1241187222

ConX: I followed the first option, renamed the files, and added aliases.

Snailedlt (approved):
LGTM! :heavy_check_mark: 
Thanks for another great contibution! :rocket:

lunatic-fox (approved):
Great work! ✔

Snailedlt (approved):

PR #1668, opened 2023-02-07
Major build improvements

Panquesito7:
Things added/changed:

- Major build improvements.
  - New detailed guide on how to build Devicon locally has been added (Windows & Linux-based systems).
  - Building Devicon on [Gitpod.io](https://www.gitpod.io) is now possible with a single click.
  - Sneak peek of new icons can be seen by building locally.
  - A few other improvements were done, such as providing a Geckodriver ([latest version](https://github.com/mozilla/geckodriver/releases/tag/v0.32.1)) for both Linux 64 and Windows 64.
  - The website will now check if CSS/build files are found in the root directory, otherwise, will use predefined [CDN](https://www.jsdelivr.com/package/npm/devicon) latest release links.
  - P.S.: a few of the file changes (such as the `font` folder changes or `devicon-base.css`) weren't changed. Not sure why GitHub displays it like that. 🤔

**NOTE:** This will need proper testing. I already tested and it works perfectly fine without any issues.
Any doubts, suggestions, or questions, please let me know about it. Thanks! 🙂

[Gitpod ready-to-code](https://gitpod.io/#https://github.com/devicons/devicon/pull/1668)


README.md
```diff
@@ -203,34 +203,119 @@ All official releases shall be in <code>master</code>. Any updates in between (i
...
+    <li><a href="#build-css">Build the CSS stylesheet</a></li>
+    <li><a href="#web-server">Setting up the web server</a></li>
+</ol>
+
+<h3 id="using-gitpod">Using Gitpod.io</h3>
+<p>By using <a href=https://www.gitpod.io)>Gitpod.io</a>, you can easily build the icons and install the<br>required dependencies in one single click. No extra setup is required.</p>
+
+<a href=https://gitpod.io/#https://github.com/devicons/devicon/tree/develop><img src=https://gitpod.io/button/open-in-gitpod.svg alt="Open in Gitpod"></img></a>
+
+<h3 id="installation">Installation</h3>
```
Snailedlt, line 225: I suggest specifying that it's a local installation to better differentiate Gitpod and local installation. Maybe we should also have the other points as subpoints to the local installs, since (if I understand it correctly) they are done automatically when using gitpod.
```suggestion
    <li><a href="#local-installation">Local Installation (Linux/Windows)</a></li>
    <ol>
      <li><a href="#building-icons">Build the icons</a></li>
      <li><a href="#build-css">Build the CSS stylesheet</a></li>
      <li><a href="#web-server">Setting up the web server</a></li>
    </ol>
</ol>

<h3 id="using-gitpod">Using Gitpod.io</h3>
<p>By using <a href=https://www.gitpod.io)>Gitpod.io</a>, you can easily build the icons and install the<br>required dependencies in one single click. No extra setup is required.</p>

<a href=https://gitpod.io/#https://github.com/devicons/devicon/tree/develop><img src=https://gitpod.io/button/open-in-gitpod.svg alt="Open in Gitpod"></img></a>

<h3 id="local-installation">Local Installation</h3>
```

README.md
```diff
@@ -203,34 +203,119 @@ All official releases shall be in <code>master</code>. Any updates in between (i
...
+
+<p><a href=https://github.com/devicons/devicon/fork>Fork</a> the repository and clone the forked repository.</p>
+
+```bash
+git clone https://github.com/<your-github-username>/devicon.git
+```
+
+<p>In case you don't have Git installed, check the <a href="https://git-scm.com/book/en/v2/Getting-Started-Installing-Git">official guide</a> to install Git on your operating system.</p>
```
Snailedlt, line 233: Perhaps we can add the "note" section for stuff like this?
ref: https://github.com/community/community/discussions/16925

Not sure how it acts together with html though, and I couldn't find the html alternative to it either :/

Anyways, if you agree, we can test this out, and potentially add it other places too:

```suggestion
> **Note**
> In case you don't have Git installed, check the <a href="https://git-scm.com/book/en/v2/Getting-Started-Installing-Git">official guide</a> to install Git on your operating system.
```

README.md
```diff
@@ -203,34 +203,119 @@ All official releases shall be in <code>master</code>. Any updates in between (i
...
+
+<h4>Install all the necessary NPM dependencies</h4>
+
+<p>In case you don't have NPM installed, check this <a href=https://kinsta.com/blog/how-to-install-node-js/><b>ultimate guide</b></a> on installing Node.js and NPM. These tools are required in order to build Devicon properly.<p>
```
Snailedlt, line 237: This can also be added as a note, see my previous comment

README.md
```diff
@@ -203,34 +203,119 @@ All official releases shall be in <code>master</code>. Any updates in between (i
...
 
 ```bash
 npm install
 ```
 
-<h5>Build the font and export stylesheet</h5>
-Open <a href="https://icomoon.io/app/#/select">icomoon.io</a> and import <a href="/icomoon.json">icomoon.json</a>. Choose <i>yes</i> when being asked
-if you would like to restore the settings stored in the configuration file.
+<h4>Installing Mozilla Firefox</h4>
```
Snailedlt, line 243: Is mozilla firefox needed? Won't it work with Chrome for example?

README.md
```diff
@@ -203,34 +203,119 @@ All official releases shall be in <code>master</code>. Any updates in between (i
...
-    by yourself.
+Follow these steps to build the website and icons either locally or using <a href=https://www.gitpod.io>Gitpod.io</a>.
 </p>
-<h5>Prerequisites</h5>
-<p>Install <code>gulp</code> (and <code>gulp</code> plugins)</p>
+
+<h3>Table of contents</h3>
+<ol>
+    <li><a href="#using-gitpod">Using Gitpod to build</a></li>
+    <li><a href="#installation">Installation (Linux/Windows)</a></li>
```
Snailedlt, line 214: Looks like the Linux and Windows installation processes are quite different, perhaps we should either just point to the official installation websites, or split this into two sections.

I would prefer if we just pointed to the installation site, since that would require less maintenance work from us in case the installation method changes. It could for example be as simple as:

### Install Python 3.8
https://www.python.org/downloads/

### Install Firefox
https://www.mozilla.org/en-US/firefox/new/

### Install Selenium
```bash
python3 -m pip install --upgrade pip && pip install selenium==4.1.0 requests==2.25.1
```

Snailedlt (changes requested):
I haven't reviewed everything yet, but I don't have more time today, so I figured I'd post this in case you want to look at it before I get the chance to review the rest :)

icons/nano/nano-original-wordmark.svg
```diff
@@ -1 +1 @@
....37 0 0 0 .271.101.378.378 0 0 0 .272-.098.344.344 0 0 0 .102-.261.341.341 0 0 0-.104-.258.376.376 0 0 0-.27-.099c-.109 0-.199.032-.268.097s-.105.15-.105.254zm.015-.913a.33.33 0 0 0 .1.249.354.354 0 0 0 .257.097.351.351 0 0 0 .258-.098.334.334 0 0 0 .1-.251.323.323 0 0 0-.1-.244.358.358 0 0 0-.258-.096.358.358 0 0 0-.258.096.33.33 0 0 0-.099.247zm1.935.449a.559.559 0 0 1-.224-.186.464.464 0 0 1-.075-.263c0-.159.057-.289.17-.391a.626.626 0 0 1 .435-.153c.174 0 .318.051.434.153a.48.48 0 0 1 .173.375.503.503 0 0 1-.074.278.534.534 0 0 1-.228.187.519.519 0 0 1 .238.185.496.496 0 0 1 .083.286.533.533 0 0 1-.174.412c-.116.106-.267.159-.452.159s-.335-.054-.451-.161a.54.54 0 0 1-.173-.417.484.484 0 0 1 .318-.464zm-.066.464c0 .108.034.196.102.264a.37.37 0 0 0 .271.101.378.378 0 0 0 .272-.098.344.344 0 0 0 .102-.261.341.341 0 0 0-.104-.258.376.376 0 0 0-.27-.099.376.376 0 0 0-.268.097.331.331 0 0 0-.105.254zm.015-.913a.33.33 0 0 0 .1.249.354.354 0 0 0 .257.097.351.351 0 0 0 .258-.098.334.334 0 0 0 .1-.251.323.323 0 0 0-.1-.244c-.067-.064-.153-.096-.258-.096s-.191.032-.258.096a.334.334 0 0 0-.099.247zm1.846.244a.555.555 0 0 1 .215-.164.713.713 0 0 1 .291-.057.703.703 0 0 1 .719.722c0 .201-.07
```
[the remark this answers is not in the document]
Panquesito7: > P.S.: a few of the file changes (such as the `font` folder changes or `devicon-base.css`) weren't changed. Not sure why GitHub displays it like that. 🤔

package.json
```diff
@@ -5,10 +5,12 @@
   "main": "devicon.min.css",
   "scripts": {
     "build-css": "gulp updateCss && gulp clean",
-    "peek-test": "python ./.github/scripts/icomoon_peek.py ./.github/scripts/build_assets/geckodriver-v0.27.0-win64/geckodriver.exe ./icomoon.json ./devicon.json ./icons ./  --pr_title \"%PR_TITLE%\"",
-    "build-test": "python ./.github/scripts/icomoon_build.py ./.github/scripts/build_assets/geckodriver-v0.27.0-win64/geckodriver.exe ./icomoon.json ./devicon.json ./icons ./",
+    "peek-test": "python3 ./.github/scripts/icomoon_peek.py ./.github/scripts/build_assets/geckodriver-v0.32.1-linux64/geckodriver ./icomoon.json ./devicon.json ./icons ./  --pr_title \"%PR_TITLE%\"",
+    "build-test": "python3 ./.github/scripts/icomoon_build.py ./.github/scripts/build_assets/geckodriver-v0.32.1-linux64/geckodriver ./icomoon.json ./devicon.json ./icons ./",
```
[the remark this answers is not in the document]
Panquesito7: Not sure what do you mean by that? 🤔

package.json
```diff
@@ -28,10 +30,13 @@
...
+    "gulp-footer": "^2.0.2",
     "gulp-sass": "^5.0.0",
     "gulp-svgmin": "^3.0.0",
     "sass": "^1.26.10",
-    "yargs": "^17.0.0",
-    "gulp-footer": "^2.0.2"
+    "yargs": "^17.0.0"
+  },
+  "dependencies": {
+    "selenium": "^2.20.0"
```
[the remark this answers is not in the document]
Panquesito7: Not sure about that. I think I installed it by accident while doing lots of tests. Will delete. Thanks. 👍

.github/scripts/build_assets/arg_getters.py
```diff
@@ -36,7 +36,7 @@ def get_selenium_runner_args(peek_mode=False):
     if peek_mode:
         parser.add_argument("pr_title",
                             help="The title of the PR that we are peeking at")
-    else:
+    elif has_token != False:
```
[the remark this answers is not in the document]
Panquesito7: > Also, if one excludes the other, why do we need two variables?

Could you explain this, please? Not 100% sure what you mean.

> what happens if `has_token` and `peek_mode` is false?
> Seems like the method won't return anything then, will that cause issues?

Good question. I'll add the token on an `else` statement, as the token is required, while the PR title is not, I think.

.gitpod.yml
```diff
@@ -0,0 +1,22 @@
+image:
+  file: .gitpod.dockerfile
+
+tasks:
+  - name: Setup & build
+    init: chmod +x ./.github/scripts/build_assets/geckodriver-v0.32.1-linux64/geckodriver
```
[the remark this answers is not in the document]
Panquesito7: I don't really think it matters much? That's being run only for Gitpod.

.gitpod.yml
```diff
@@ -0,0 +1,22 @@
...
+    command: npm install && npm run build-icons
+  - name: Build CSS & run web server
+    init: npm run build-css && npm run dev
```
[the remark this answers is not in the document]
Panquesito7: Yeah. That extension is no longer needed.

README.md
Panquesito7: I like the idea! I'll see how can we do that in HTML. If not, I'll use Markdown. Thanks. 👍

icons/nano/nano-original-wordmark.svg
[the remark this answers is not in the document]
Snailedlt: If they weren't changed, could you please revert them so we don't make any accidental changes? :)

.github/scripts/build_assets/arg_getters.py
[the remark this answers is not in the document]
Snailedlt: > > Also, if one excludes the other, why do we need two variables?
>
> Could you explain this, please? Not 100% sure what you mean.

What is the `peek_mode` variable for? Seems to me like it only does the opposite of `has_token`. If it's true, then the PR title argument will be added, if it's not true, then the token argument will be added. If `has_token` is true, then the opposite will happen.

Shouldn't they either be unrelated (e.g. two if statements instead of if-else), or just be one variable?

README.md
Snailedlt: Markdown is easier to understand and shorter to write, so if we can transition to markdown in the readme, that would be great too :)

Panquesito7: > Markdown is easier to understand and shorter to write, so if we can transition to markdown in the readme, that would be great too :)

Yeah. I was thinking to do the same, too. We can do that in another PR. 🙂

icons/nano/nano-original-wordmark.svg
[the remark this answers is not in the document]
Panquesito7: > If they weren't changed, could you please revert them so we don't make any accidental changes? :)

The changes for the `nano` icons are reverted.
I think the other files aren't even changed, probably just a Git/GitHub-related thing.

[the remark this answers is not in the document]
Snailedlt: @Panquesito7 There's still one nano file which has been changed in this PR

README.md
Snailedlt: bump

Panquesito7: > Is mozilla firefox needed? Won't it work with Chrome for example?

Nope, Mozilla Firefox is needed for this.
I tried it without Firefox, but it didn't work.

icons/nano/nano-original-wordmark.svg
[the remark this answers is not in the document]
Panquesito7: Strange. I didn't change anything. Probably some endline stuff.

.github/scripts/build_assets/arg_getters.py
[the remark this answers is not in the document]
Panquesito7: I edited the message by the way.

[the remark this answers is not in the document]
Snailedlt: Wouldn't it be better to change it from elseif to just an if then?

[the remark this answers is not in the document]
Snailedlt: Can't it just be as simple as this?
```suggestion
    if has_token:
```

lunatic-fox (approved):
Looks good to me! 🚀

icons/nano/nano-original-wordmark.svg
Snailedlt, line 1: This change still needs to be reverted

.github/scripts/build_assets/arg_getters.py
```diff
@@ -36,7 +36,7 @@ def get_selenium_runner_args(peek_mode=False):
     if peek_mode:
         parser.add_argument("pr_title",
                             help="The title of the PR that we are peeking at")
-    else:
+    if has_token != False:
```
Snailedlt, line 39: This is a bit easier on the eyes imo :)
```suggestion
    if not has_token:
```

It's preference ofc, so feel free to ignore this if you disagree

docs/index.html
```diff
@@ -31,19 +31,76 @@
...
+
+          if (counter == 1) {
+            clearInterval(timer);
+          }
+        }, 0);
+  </script>
 </head>
 
 <body ng-app="devicon" ng-controller="IconListCtrl">
-  <link rel="stylesheet" ng-if="latestReleaseTagging" ng-href="https://cdn.jsdelivr.net/gh/devicons/devicon@{{ latestReleaseTagging }}/devicon.min.css">
```
Snailedlt, line 45: Is this not needed anymore? Could you explain what it did, and why we don't need it anymore?

docs/index.html
```diff
@@ -88,7 +145,7 @@ <h4>
             </span>
           </p>
           <div class="cde" id='headerLinkCode'>
-            &lt;link rel="stylesheet" href="https://cdn.jsdelivr.net/gh/devicons/devicon@{{ latestReleaseTagging }}/devicon.min.css"&gt;
+            &lt;link rel="stylesheet" href="https://cdn.jsdelivr.net/gh/devicons/devicon@latest/devicon.min.css"&gt;
```
Snailedlt, line 148: Any reason why we can't use `<` and  `>` instead of `&lt;` and `&gt;`?
Also, why did we change this to latest? Won't it cause issues when we want to test css for other releases than the latest?

docs/index.html
```diff
@@ -148,7 +209,7 @@ <h4>
             </span>
           </p>
           <div class="cde" id='imgCode'>
-            &lt;img src="https://cdn.jsdelivr.net/gh/devicons/devicon/icons/{{selectedIcon.name}}/{{selectedIcon.name}}-{{selectedIcon.svg[selectedSvgIndex]}}.svg" /&gt;<br />
+            &lt;img src="https://cdn.jsdelivr.net/gh/devicons/devicon@latest/icons/{{selectedIcon.name}}/{{selectedIcon.name}}-{{selectedIcon.svg[selectedSvgIndex]}}.svg" /&gt;<br />
```
Snailedlt, line 212: Same here, why not use `<` and `>`? Also why use `@latest`, does it provide any benefit?

docs/index.html
```diff
@@ -31,19 +31,76 @@
...
+    var timer = setInterval(function() {
+          ++counter;
+
+          addCss(`${baseUrl}devicon.min.css`);
+          addImage(`${baseUrl}icons/{{selectedIcon.name}}/{{selectedIcon.name}}-{{svgVersion}}.svg`, "svg-list", 'svgLogo');
+
+          if (counter == 1) {
+            clearInterval(timer);
+          }
+        }, 0);
```
Snailedlt, line 99: I'm a bit confused. Could you tell me what this code does?

Snailedlt (changes requested):
README looks good now 💯 

One of the nano icons is still changed and needs to be reverted though.
Other than that I mostly just have some questions to better understand the code.

Not much left now until we can merge 🚀

icons/nano/nano-original-wordmark.svg
Panquesito7: I reverted all the changes in the icon; not sure what else should I do. 🤔

docs/index.html
Panquesito7: 1. Yes it is needed, but it is added by the script above.
2. The stylesheet file (`devicon.min.css`) loads all the icons data, colors, etc..
3. As I said, it is added by the JavaScript script above.

docs/index.html
Panquesito7: 1. Not sure what the difference between those two is. I could try, but I'm not sure it will benefit anything.
2. Latest will always fetch the latest **released version** in NPM (in this case, 2.15.1), so it is the same as specifying the version manually or using `@latest`, which is faster, though. No errors should happen, I already tested it and it works fine.

docs/index.html
Panquesito7: Already commented above about the same.

docs/index.html
Panquesito7: This will add the `devicon.min.css` stylesheet in the HTML head, however, we must verify whether the website is being built locally or using the NPM provided files, that's why this is done on this script.
The image addition is adding the selected icon, but we must determine whether local or NPM files should be used (see the `svgLogo` ID in one of the HTML `div`s for more information).

The counter interval would usually be useful to add multiple instances of those (we just need one here, though). However, when I tried that without the interval, it didn't add them properly or work, so I used that as a workaround.

docs/index.html
Snailedlt: 1. Should be the same (I think), but it looks weird with `&lt;` instead of `<`
2. Hmm, what if we made a breaking change, and we needed to see the difference between latest and say version 2.15.0. wouldnt it be easier if `@latest` wasn't hardcoded?

docs/index.html
Snailedlt: I see. The need for a timer seems odd, any idea why it works with the timer, but not without it?

Panquesito7: > I see. The need for a timer seems odd, any idea why it works with the timer, but not without it?

Nope, I don't have a clue why that happens. 🤔

icons/nano/nano-original-wordmark.svg
Snailedlt: File history: https://github.com/devicons/devicon/commits/f8ecce93a4ceb7dd3fb93ab6c7343f2fe458e532/icons/nano/nano-original-wordmark.svg

docs/index.html
Snailedlt: Oh well, if it works... It works 😅

docs/index.html
Snailedlt: I see. It would probably be less code to move the script to the .js file and add a variable, then use that variable here instead of adding the element via the DOM.
That would more closely follow the SOLID principle too

docs/index.html
Snailedlt: Actually, can we move this to the index.js file instead? Then we keep the script separate from the html, like before.

If you allow me to edit your branch I can probably work on it tomorrow :)

icons/nano/nano-original-wordmark.svg
Panquesito7: All I did is copy the `nano` icon contents from the `develop` branch. Probably some Git-meta changes or such.

docs/index.html
Panquesito7: So, you're saying to move the whole script to another JS file and load the script via HTML?

docs/index.html
Panquesito7: > difference between latest and say version 2.15.0

What do you mean exactly? 🤔

Snailedlt: What if release 15.2.0 brought a breaking change.
To debug we might want to check out the 15.1.0 release. If `@latewt` is hardcoded like it is now, wouldn't we beed to change to `15.1.0` everywhere, instead of just one place (like before)?

Panquesito7: That's a good point. I tried hardcoding it like that before, but for some reason it didn't work.
I could try again if wanted, or you could try it yourself (let me know what works best for you).

docs/index.html
Snailedlt: Yes, sort of. Moving it to the .js file, but using Angular features instead of interacting directly with the DOM. It's less complex and less code that way.

icons/nano/nano-original-wordmark.svg
Snailedlt: Ahhh, have you tried `git revert`?

docs/index.html
Snailedlt: Feel free to try if you'd like, I'll also take a look ASAP (which is hopefully tomorrow)

lunatic-fox: `&lt;` and `&gt;` are html escape sequences to literal characters `<` and `>`. So if we want to show a string with those characters on html (hardcoded) we need to use the escape sequences, otherwise the string will be interpreted as a code.
A workaround can be achieved by inserting this text via JavaScript, like:

***index.html***
```html
<div class="cde" id='headerLinkCode'></div>
```
***./assets/js/script.js***
```js
const headerLinkCodeElem = document.getElementById('headerLinkCode');
headerLinkCodeElem.textContent = `<link rel="stylesheet"href="https://cdn.jsdelivr.net/gh/devicons/devicon@latest/devicon.min.css">`;
```
However, I'm ok as it is now. 😁

I agree with @Snailedlt on not hardcode `@latest`, for the same reasons.

Snailedlt: Just to make it absolutely clear. 
`&lt;` and `&gt;` are used here, because we want to show `<` and `>` inside these fields (which means we need to use the escape-sequence instead of the actual character symbol): 
image

Snailedlt: I changed back to variable version, and moved the JS code to the .ja file in this PR: https://github.com/Panquesito7/devicon/pull/187

@lunatic-fox feel free to take a look as well. I wasn't able to get the local SVG's working on my and panquesito's branch, so if you could test it locally, that would be neat!

lunatic-fox: Exactly! 🙂

Snailedlt:
> 

Thanks. It works perfectly with those fonts!
image

How did you build it?

lunatic-fox:
> @lunatic-fox
> 
> Thanks. It works perfectly with those fonts! image
> 
> How did you build it, and which OS do you use?

I made a script with Node.js to get all valid icons to upload to Icomoon and I'm on Windows 10.

Snailedlt:
@lunatic-fox care to share the script? :)

lunatic-fox:
> > @lunatic-fox care to share the script? :)
> 
> I made it first with local files, but I've made another version using fetch. This will need a `icomoon` directory created and `Node.js 17.5.^`.
> 

Here goes a faster version:
```js
const fs = require('fs');
const baseURL = 'https://cdn.jsdelivr.net/gh/devicons/devicon@develop/';
(async() => {
  let x = 0, y = 10;
  const devicon = await(await fetch(`${baseURL}devicon.json`)).json();
  const toBuild = [];

  while (true) {
    const deviconChunk = devicon.map(e => ({
      name: e.name,
      font: e.versions.font
    })).slice(x, y);

    if (deviconChunk.length === 0) break;

    for (const e of deviconChunk) {
      for (const f of e.font) {
        const icon = await fetch(`${baseURL}icons/${e.name}/${e.name}-${f}.svg`);
  
        if (icon.status === 200) {
          const svg = await icon.text();
          toBuild.push({
            path: `${e.name}-${f}.svg`,
            icon: svg
          });
          console.log(`> ${e.name}-${f}`);
        }
      }
    }
    x += 10, y += 10;
  }

  toBuild.forEach(e => fs.writeFileSync(`icomoon/${e.path}`, e.icon));
})();
```

Snailedlt (approved):
Looks good! ✔️ 

Like we already discussed, I had trouble running it locally. If the issue persists we can look at that later. For now it's good enough that it runs on GitPod imo :)
Let's get this release out ASAP :D

PR #1663, opened 2023-02-03
Update and use Geckodriver for Linux64

Panquesito7:
Things added/changed:

- Update and use Geckodriver for Linux64.
  - We might want to make sure the newest Geckodriver version works. If it doesn't, we can always revert back to the other version.
  - The Windows build is kept in the `build_assets` folder in case it's needed again. If wanted, we can delete it.

PR #1675, opened 2023-02-10
Update and use Geckodriver for Linux64

Panquesito7:
Things added/changed:

- Update and use Geckodriver for Linux64.
  - Does the same as #1663, but towards `develop`.


.github/workflows/build_icons.yml
```diff
@@ -18,13 +17,12 @@ jobs:
...
       - name: Executing build and create fonts via icomoon
-        shell: cmd
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: >
           python ./.github/scripts/icomoon_build.py 
-          ./.github/scripts/build_assets/geckodriver-v0.30.0-win64/geckodriver.exe ./icomoon.json 
-          ./devicon.json ./icons ./ %GITHUB_TOKEN% --headless
+          ./.github/scripts/build_assets/geckodriver-v0.32.2-linux64/geckodriver ./icomoon.json 
+          ./devicon.json ./icons ./ "$GITHUB_TOKEN" --headless
```
Snailedlt, line 25: Did this fix the issue with the github token not working? If so, is it because the syntax for referring to a variable is different in bash and cmd?

package.json
```diff
@@ -5,9 +5,8 @@
   "main": "devicon.min.css",
   "scripts": {
     "build-css": "gulp updateCss && gulp clean",
-    "peek-test": "python ./.github/scripts/icomoon_peek.py ./.github/scripts/build_assets/geckodriver-v0.27.0-win64/geckodriver.exe ./icomoon.json ./devicon.json ./icons ./  --pr_title \"%PR_TITLE%\"",
-    "build-test": "python ./.github/scripts/icomoon_build.py ./.github/scripts/build_assets/geckodriver-v0.27.0-win64/geckodriver.exe ./icomoon.json ./devicon.json ./icons ./",
-    "optimize-svg": "gulp optimizeSvg",
+    "peek-test": "python ./.github/scripts/icomoon_peek.py ./.github/scripts/build_assets/geckodriver-v0.32.2-linux64/geckodriver ./icomoon.json ./devicon.json ./icons ./  --pr_title \"%PR_TITLE%\"",
```
Snailedlt, line 8: Don't we need to replace `%PR_TITLE%` with `$PR_TITLE` here too?

package.json
```diff
@@ -5,9 +5,8 @@
   "main": "devicon.min.css",
   "scripts": {
     "build-css": "gulp updateCss && gulp clean",
-    "peek-test": "python ./.github/scripts/icomoon_peek.py ./.github/scripts/build_assets/geckodriver-v0.27.0-win64/geckodriver.exe ./icomoon.json ./devicon.json ./icons ./  --pr_title \"%PR_TITLE%\"",
-    "build-test": "python ./.github/scripts/icomoon_build.py ./.github/scripts/build_assets/geckodriver-v0.27.0-win64/geckodriver.exe ./icomoon.json ./devicon.json ./icons ./",
-    "optimize-svg": "gulp optimizeSvg",
```
Snailedlt, line 10: Why was `optimize-svg` removed?

Snailedlt (changes requested):
Just some questions

.github/workflows/build_icons.yml
Panquesito7: Yeah, it's different. And yes, it's fixed.

package.json
Panquesito7: You're right, we need that. I'll change that immediately.

package.json
Panquesito7: Whoops, removed that by accident. I wanted to remove `build-test`, as that one doesn't work. It asks for a GitHub token. PR #1668 adds a command to build the icons locally/Gitpod.

Snailedlt (approved):
Assuming it's been tested, I think this is safe to approve now 👍
Thanks for the PR @Panquesito7

lunatic-fox (approved):

PR #1695, opened 2023-03-04
add feature: newliner

### LINKED ISSUE #1624: [FEATURE REQUEST] Automatically add newline to all files in new PRs

- [X] Confirm

### Problem

There's often a missing newline at the end of the files, which [can cause issues with old code editors.](https://stackoverflow.com/a/2287990/12206312)

### Possible Solution

Add the [newline-action](https://github.com/marketplace/actions/newline-action) to a new github workflow

### Additional information

This will up the quality of devicons, and boost productivity for both reviewers and code contributors.

lunatic-fox:

## Features
This feature adds `newliner.py` and changes `optimize_icons.yml` to `optimizer.yml`.

### About `optimizer`
`optimizer` workflow is triggered by any user commit to the pull request (synchronize) and by `feature:icon` label.

- `synchronize` - This trigger adds a newline at the end of each file that is missing it.
- `feature:icon` - This trigger optimizes any icon with SVGO, it also adds a newline at the end of the optimized icon.

**This PR closes #1624 and changes #1554**

## Notes
We can change the label that triggers the icon optimization.

I've decided to make these two actions in one workflow, because some triggers are conflicting or not working properly, but if some way can be found to split them it would be nice to apply it.


.github/workflows/optimizer.yml
```diff
@@ -0,0 +1,48 @@
...
+      - name: Save paths added and modified files
+        run: echo '${{ steps.pr-files.outputs.added_modified }}' > pr-files.json
+
+      - name: Setup Python v3.10
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.10'
+
+      - name: Add newline
+        run: python ./.github/scripts/newliner.py ${{ env.SVGO }}
```
Snailedlt, line 44: Any reason why we can't use the [newline-action](https://github.com/marketplace/actions/newline-action)?
```suggestion
      - name: Add newline
        uses: Logerfo/newline-action@0.0.4
        with:
          github-token: ${{ secrets.GITHUB_TOKEN }}
```

This would add newline to any file, not just SVG files. 
It's also possible to ignore files in a config (see link above)

Come to think of it, maybe it would be better to add the newline in a separate script like this:

```yml
name: Add newline
on:
  pull_request:
    types: [synchronize, opened]
jobs:
  build:
    name: newline-action
    runs-on: ubuntu-16.04
    steps:
      - uses: actions/checkout@master
      - uses: Logerfo/newline-action@0.0.4
        with:
          github-token: ${{ secrets.GITHUB_TOKEN }} # The `GITHUB_TOKEN` secret.
          config-path: .github/newline.yml # The path of the addtional configurations file.
```

Snailedlt (changes requested):
I suggest we use [newline-action](https://github.com/marketplace/actions/newline-action) instead of making our own newline script. And also I would prefer it if the script was moved to a different workflow all together. 

See comment below for more info

.github/workflows/optimizer.yml
Panquesito7: This sounds good to me. 👍

canaleal:
Has not been updated in 2 years. Closing, but can be reopened.

PR #1724, opened 2023-04-10
Bugfix: Remove Optimize Icons workflow from develop branch

### LINKED ISSUE #1715: [BUG] Optimize icons with SVGO check fails

lunatic-fox:

## Features
This pull request removes `Optimize Icons` workflow from `develop` as a temporary fix, since that workflow keeps failing due to permission issues.

**This PR closes #1715, #1718**


Panquesito7 (approved):
Thanks. Let's try to fix out the error as soon as possible (count me in). 🙂

Snailedlt (approved):

PR #1731, opened 2023-04-12
Update many of the actions to their latest version

Panquesito7:
Things added/changed:

- Update many of the actions to their latest version.
  - **NOTE:** I have not tested the workflows. They should be fully tested before merging.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,62 +1,61 @@
-name: "CodeQL"
+name: "Code Scanning - Action"
 
 on:
   push:
     branches: [master]
   pull_request:
-    # The branches below must be a subset of the branches above
-    branches: [master]
+    branches: [master, develop]
   schedule:
-    - cron: '0 23 * * 4'
+    #        ┌───────────── minute (0 - 59)
+    #        │  ┌───────────── hour (0 - 23)
+    #        │  │ ┌───────────── day of the month (1 - 31)
+    #        │  │ │ ┌───────────── month (1 - 12 or JAN-DEC)
+    #        │  │ │ │ ┌───────────── day of the week (0 - 6 or SUN-SAT)
+    #        │  │ │ │ │
+    #        │  │ │ │ │
+    #        │  │ │ │ │
+    #        *  * * * *
+    - cron: '30 1 * * 0'
 
 jobs:
-  analyze:
-    name: Analyze
+  CodeQL-Build:
+    # CodeQL runs on ubuntu-latest, windows-latest, and macos-latest
     runs-on: ubuntu-latest
 
-    strategy:
-      fail-fast: false
-      matrix:
-        # Override automatic language detection by changing the below list
-        # Supported options are ['csharp', 'cpp', 'go', 'java', 'javascript', 'python']
-        language: ['javascript']
-        # Learn more...
-        # https://docs.github.com/en/github/finding-security-vulnerabilities-and-errors-in-your-code/configuring-code-scanning#overriding-automatic-language-detection
+    permissions:
+      # required for all workflows
+      security-events: write
+
+      # only required for workflows in private repositories
+      actions: read
+      contents: read
 
     steps:
-    - name: Checkout repository
-      uses: actions/checkout@v2
-      with:
-        # We must fetch at least the immediate parents so that if this is
-        # a pull request then we can checkout the head.
-        fetch-depth: 2
+      - name: Checkout repository
+        uses: actions/checkout@v3
 
-    # If this run was triggered by a pull request event, then checkout
-    # the head of the pull request instead of the merge commit.
-    - run: git checkout HEAD^2
-      if: ${{ github.event_name == 'pull_request' }}
+      # Initializes the CodeQL tools for scanning.
+      - name: Initialize CodeQL
+        uses: github/codeql-action/init@v2
+        # Override language selection by uncommenting this and choosing your languages
+        # with:
+        #   languages: go, javascript, csharp, python, cpp, java, ruby
 
-    # Initializes the CodeQL tools for scanning.
-    - name: Initialize CodeQL
-      uses: github/codeql-action/init@v1
-      with:
-        languages: ${{ matrix.language }}
+      # Autobuild attempts to build any compiled languages (C/C++, C#, Go, or Java).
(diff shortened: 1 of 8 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -1,62 +1,61 @@
-name: "CodeQL"
+name: "Code Scanning - Action"
 
 on:
   push:
     branches: [master]
```
Snailedlt, line 5: Shouldn't `develop` also be added here? I have no idea myself, just thought so based on the old comment:
image

.github/drafts/optimize_svg.yml
```diff
@@ -6,7 +6,7 @@ jobs:
   peek:
     name: Optimize the added/changed svgs
     if: github.event.label.name == 'bot:optimize'
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-latest18.04
```
Snailedlt, line 9:
```suggestion
    runs-on: ubuntu-latest
```

.github/drafts/stale.yml
```diff
@@ -5,7 +5,7 @@ on:
 
 jobs:
   stale:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-latestlatest
```
Snailedlt, line 8:
```suggestion
    runs-on: ubuntu-latest

File: .github/workflows/peek_icons.yml [hunk 4]
```diff
@@ -9,12 +9,12 @@ jobs:
     # fail due to system, skipped
     name: Peek Icons
     if: github.event.label.name == 'bot:peek'
-    runs-on: windows-2019
+    runs-on: windows-latest
```
Snailedlt, line 12: Wait, why does this run on windows?

Snailedlt (changes requested):

.github/workflows/peek_icons.yml
Panquesito7: It was due to the Geckodriver being used in Windows only.
We can change it to Linux, but I don't see any difference at all.

Snailedlt: Linux is faster, more stable, more frequently updated and more storage efficient.
So if we can use it, we should :)

Panquesito7: Changed to use `ubuntu-latest` as well as fixed an issue with the workflow.

Panquesito7:
I've tested a few workflows on my fork and they seem to be working fine.
Please check again and make sure everything's correct. Thanks. 🙂

.github/workflows/in_develop_labeler.yml
```diff
@@ -8,7 +8,7 @@ jobs:
   label_preflight:
     name: Label Issue In Develop
     runs-on: ubuntu-latest
-    if: github.event.pull_request.merged == true
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
```
Snailedlt, line 11: Nice, this is much better! 
Didn't know of the `.conclusion` event.

We should probably have a clause for failure as well, maybe we can just console log something along the lines of 
> Label Preflight workflow failed, please check the logs [Link to preflight workflow run]

So maybe something like this?

```yml
   on-failure:
     runs-on: ubuntu-latest
     if: ${{ github.event.workflow_run.conclusion == 'failure' }}
     steps:
       - run: echo "First workflow was a failure"
 ```

File: .github/workflows/in_develop_labeler.yml [hunk 6]
```diff
@@ -8,7 +8,7 @@ jobs:
   label_preflight:
```
Snailedlt, line 8:
```suggestion
  on-failure:
    runs-on: ubuntu-latest
    if: ${{ github.event.workflow_run.conclusion == 'failure' }}
    steps:
      - run: echo "First workflow was a failure"
  label_preflight:
```

Snailedlt: Dunno how to link to the workflow run, but that should be possible too if needed

Snailedlt (changes requested):

.github/workflows/peek_icons.yml
```diff
@@ -38,7 +38,6 @@ jobs:
       - name: Run icomoon_peek.py
         env:
           PR_TITLE: ${{ github.event.pull_request.title }}
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
Snailedlt: Why do we need to remove this btw?

Panquesito7: The token's not even needed. The script was not configured properly, it should be fixed now, though.

Snailedlt (approved):
Statically it looks good, and I trust you tested all of the workflows properly :)

Thanks for the contributions @Panquesito7 🙇

Panquesito7:
> Statically it looks good, and I trust you tested all of the workflows properly :)

Thank you for reviewing! In case there's anything wrong, we can always revert this PR. 😉

.github/workflows/post_peek_screenshot.yml
```diff
@@ -20,22 +20,22 @@ jobs:
         run: echo $PEEK_STATUS
 
       - name: Download workflow artifact
-        uses: dawidd6/action-download-artifact@v2.11.0
+        uses: dawidd6/action-download-artifact@v2.26.1
```
lunatic-fox, line 23: There is a recently update to that [**action**](https://github.com/dawidd6/action-download-artifact). 
```suggestion
        uses: dawidd6/action-download-artifact@v2.27.0
```

lunatic-fox (changes requested):
Great work! 🚀
Just one suggestion about `dawidd6/action-download-artifact` action version and @Snailedlt suggestion on [`.github/drafts/stale.yml`](https://github.com/devicons/devicon/pull/1731/files#r1173922783).

.github/workflows/post_peek_screenshot.yml
```diff
@@ -7,7 +7,7 @@ on:
 jobs: 
   post_screenshots_in_comment:
     name: Post the screenshot
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-latest
```
lunatic-fox, line 10: ⚠ This change will take effect when this commit is merged to `master`, as for now this action will be loading indefinitely to find a `ubuntu-18.04` machine.

PR #1764, opened 2023-05-17
Add optimize_icons workflow triggered on push

### LINKED ISSUE #1715: [BUG] Optimize icons with SVGO check fails

ConX:
## Double-check these details before you open a PR

## Features

This PR closes #1715

## Notes

This PR replaces the now closed #1718.

In the previous version, we were trying to make it run on the PR itself, but that doesn't seem to be possible, given the following:

- The `ericcornelissen/svgo-action@v3` action [does not support](https://github.com/ericcornelissen/svgo-action/blob/main/docs/events.md) a `workflow_run` trigger
- The GITHUB_TOKEN [does not have permission to write](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) when invoked by PRs from public forks

This version simply runs the optimizer after the PR merge (on the corresponding push). I have tested this with a second fork of the repository, and it seems to work as expected. According to the [documentation](https://github.com/ericcornelissen/svgo-action/blob/main/docs/events.md#on-push)) when SVGO is invoked on `push`:

> In the push context the SVGO Action will optimize all SVGs that have been added or modified in the commit(s) being pushed. Any SVGs that are in the repository but have not been modified in the commit(s) will not be optimized.

Doing this on the push has the disadvantage of not allowing us to review the optimized versions, but it will ensure all icons in `develop` are optimized.

### DIFF
--- a/.github/workflows/optimize_icons.yml
+++ b/.github/workflows/optimize_icons.yml
@@ -1,0 +1,23 @@
+name: Optimize icons
+on:
+  push:
+    branches:
+      - develop
+
+jobs:
+  optimize:
+    runs-on: ubuntu-latest
+    name: Optimize icons with SVGO
+    steps:
+      - uses: actions/checkout@v3
+
+      - uses: ericcornelissen/svgo-action@v3
+        id: svgo
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          svgo-config: ./.github/scripts/svgo.config.js
+          svgo-version: 3
+
+      - uses: stefanzweifel/git-auto-commit-action@v4
+        with:
+          commit_message: "SVG Optimization: ${{ steps.svgo.outputs.OPTIMIZED_COUNT }}/${{ steps.svgo.outputs.SVG_COUNT }} SVGs got optimized."


Panquesito7:
Not sure if the commit will be able to be pushed as the `develop` branch is protected.
Would a PR towards the `develop` branch work in this case? 🤔

Panquesito7:
Probably we should configure the action to create a PR towards the `develop` branch only.
Granting access to GitHub Actions isn't really safe. If someone modifies the actions in a PR, they could end up doing unwanted changes to the repository, AFAIK.

Finii:
> Granting access to GitHub Actions isn't really safe.

ACK ;)

> If someone modifies the actions in a PR, they could end up doing unwanted changes to the repository, AFAIK.

Well, the action modification must pass a review and merge first. Changing the action in a PR runs with the actions of the current (not the PR'ed) version. Hmm, is that easy to understand? A PR runs with the current actions, even if it wants to introduce changes with the actions. The changed actions are only used after they have been pulled.

In my opinion that makes it ok-ish to have actions with write permissions, just everyone who can merge must know this and watch out for unwanted changes :grimacing:

Snailedlt:
@Finii The github action from the main branch is ran on PRs, that's true... But I think the scripts it refers to are fetched from the current branch. So if the `./.github/scripts/svgo.config.js` is changed in a PR it will be ran in that PR.

I don't think this script in particular poses a threat as it is now... But it depends on how the svgo action runs the config script. Since it is a .js script it might be vulnerable depending on how it's interpreted

PR #1767, opened 2023-05-18
Update codeql-analysis to run on .github

ConX:
## Double-check these details before you open a PR

- x] PR does not match another non-stale PR currently opened

## Features

It appears that by default codeql does not look into the `.github` directory, and as a result, all executions fail as python is listed, but no `.py` scripts exist outside that hidden directory.

To address this, this PR splits the codeql initialization and analysis into three parallel jobs, one for javascript, one for python that runs in `.github/`, one optional python for the rest in case we later add python scripts outside the `.github/` directory.

Here are some screenshots from the testing:

### Successful execution of all codeql analysis jobs.
image

### Log that shows that the Python is skipped since there are no files there
image

**This PR closes NONE**

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -6,56 +6,80 @@ on:
   pull_request:
     branches: [master, develop]
   schedule:
-    #        ┌───────────── minute (0 - 59)
-    #        │  ┌───────────── hour (0 - 23)
-    #        │  │ ┌───────────── day of the month (1 - 31)
-    #        │  │ │ ┌───────────── month (1 - 12 or JAN-DEC)
-    #        │  │ │ │ ┌───────────── day of the week (0 - 6 or SUN-SAT)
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -6,56 +6,80 @@ on:
...
-    # CodeQL runs on ubuntu-latest, windows-latest, and macos-latest
+  CodeQL-Javascript:
     runs-on: ubuntu-latest
-
     permissions:
-      # required for all workflows
       security-events: write
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
```
Panquesito7, line 18: We probably should use `v3` here.
```suggestion
        uses: actions/checkout@v3
```

ConX: Changed it! Thanks!

Panquesito7 (approved):
Thanks! 🚀

PR #1828, opened 2023-07-13
new icon: wolfram (original, original-wordmark, plain, plain-wordmark)

Tarang74:

**This PR closes NONE**

https://company.wolfram.com/press-center/


Tarang74:
Includes wolframlanugage icon also

lunatic-fox (changes requested):
Hello there! 👋
Thank you for contributing with this project!

We need some fixes here.
Accordingly to their site Wolfram is a company and falls under a special case that we do not accept anymore: 
  - [**Wiki - What Icons Do We Accept? - Special Cases**](https://github.com/devicons/devicon/wiki/What-Icons-Do-We-Accept%3F#special-cases-see-this-discussion-for-more-details)

However, Wolfram language works like a service and is allowed as an icon here.

That said you can:
- Remove `wolfram` icons and directory.
- Keep `wolframlanguage` directory and icon.
- Add `original-wordmark`, `plain` and `plain-wordmark` versions of `wolframlanguage`.

Here goes some suggestions of the icons that need to be added:
|`original-wordmark`|`plain`|`plain-wordmark`|
|:-:|:-:|:-:|
||||

lunatic-fox (dismissed):
`wolframlanguage` icons are ok now.
Approved! ✔️
Screenshot from 2023-08-10 16-56-54
Screenshot from 2023-08-10 16-57-29

Gautam25Raj (approved):

Snailedlt (dismissed):
Everything looks good, but I'm wondering why this is called `wolframlanguage` instead of `wolfram` and having `wolframlanguage` as an altname. Wolfram seems to be the official name, so that is what we should use.
If there is no particular reason, please change to `wolfram`

lunatic-fox:
> Everything looks good, but I'm wondering why this is called `wolframlanguage` instead of `wolfram` and having `wolframlanguage` as an altname. Wolfram seems to be the official name, so that is what we should use. If there is no particular reason, please change to `wolfram`

Wolfram is a company and Wolfram Language is a service from that company, both share the same first name, but since it seems that are no other service that fits on the scope of Devicons I think it's ok to change it to `wolfram` instead and keep `wolframlanguage` as an altname. 👍🏼

Snailedlt:
@Tarang74 Let us know if you need any help :)

lunatic-fox (dismissed):

canaleal (dismissed):

canaleal (approved):

PR #1840, opened 2023-07-23
Fix build icons workflow

lunatic-fox:

## Features

**This PR closes NONE**

## Notes
Skips `imgur_step` and removes the lines related to this step on `release_message_step`. It also uses the last version of Windows, GitHub Checkout, Setup Python and Python on 3.10 version.

### DIFF
--- a/.github/workflows/build_icons.yml
+++ b/.github/workflows/build_icons.yml
@@ -3,13 +3,13 @@ on: workflow_dispatch
 jobs:
   build:
     name: Get Fonts From Icomoon
-    runs-on: windows-2019
+    runs-on: windows-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
 
-      - uses: actions/setup-python@v2
+      - uses: actions/setup-python@v4
         with: 
-          python-version: 3.8
+          python-version: '3.10'
 
       - name: Install dependencies (python, pip, npm)
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/build_icons.yml
```diff
@@ -44,15 +44,15 @@ jobs:
...
-          client_id: ${{secrets.IMGUR_CLIENT_ID}}
+      # - name: Upload screenshot of the newly made icons
+      #   id: imgur_step
+      #   uses: devicons/public-upload-to-imgur@v2.2.2
+      #   if: success()
+      #   with:
+      #     # will have "new_icons.png" and "new_svgs.png"
+      #     # in that order (cause sorted alphabetically)
+      #     path: ./screenshots/*.png
+      #     client_id: ${{secrets.IMGUR_CLIENT_ID}}
```
lunatic-fox, line 55: This part is skipped by comment, but it can be removed.

Panquesito7: Let's keep it as-is so that we can fix it in the future. 👍

Panquesito7 (commented):
Are you sure we do not need to update other workflows to use newer versions, as done in #1731?

lunatic-fox:
> Are you sure we do not need to update other workflows to use newer versions, as done in #1731?

Well, I do think so, since this action uses geckodriver on a Windows machine and on `develop` branch geckodriver is on an Ubuntu machine.

.github/workflows/build_icons.yml
```diff
@@ -3,13 +3,13 @@ on: workflow_dispatch
 jobs:
   build:
     name: Get Fonts From Icomoon
-    runs-on: windows-2019
+    runs-on: windows-latest
```
lunatic-fox, line 6:
```suggestion
    runs-on: ubuntu-latest
```

Panquesito7 (approved):
Thanks! 🚀
Let's test it out.

PR #1956, opened 2023-10-30
New icon: Twilio (original, original-wordmark)

dariakiseleva:

**This PR closes NONE**

https://www.twilio.com/


devicon.json
```diff
@@ -9936,6 +9936,27 @@
...
+                "original",
+                "original-wordmark"
+            ],
+            "font": [
+                "original",
+                "original-wordmark"
+            ]
+        },
+        "color": "#f22f46",
+        "aliases": []
```
Snailedlt, line 9958:
```suggestion
        "aliases": [
            {
                "base": "original",
                "alias": "plain"
            },
            {
                "base": "original-wordmark",
                "alias": "plain-wordmark"
            }
        ]
```

Snailedlt (changes requested):
Hi, thank you for the PR! :D
The icons look good, and the devicon.json is mostly good as well.
It seems like twilio is the company, and that they have multiple different products... in which case it would be out of scope for devicons.
But if twilio is also a platform I do believe it's within scope. You can read more about which icons are accepted here: https://github.com/devicons/devicon/wiki/What-Icons-Do-We-Accept%3F

@dariakiseleva Could you provide some more info (perhaps some links) to confirm that twilio is within scope of devicons?

canaleal:
> Hi, thank you for the PR! :D The icons look good, and the devicon.json is mostly good as well. It seems like twilio is the company, and that they have multiple different products... in which case it would be out of scope for devicons. But if twilio is also a platform I do believe it's within scope. You can read more about which icons are accepted here: https://github.com/devicons/devicon/wiki/What-Icons-Do-We-Accept%3F
> 
> @dariakiseleva Could you provide some more info (perhaps some links) to confirm that twilio is within scope of devicons?

Looks like twilio is a platform. 

The messaging, voice, email, lookup, verify, etc tools make use of the twilio logo.

Snailedlt:
Then it's fine by me. Thanks for checking up on this again @canaleal!

canaleal (approved):

PR #2001, opened 2023-11-13
Add alias and remove duplicate files

weh:
## Double check these details before you open a PR

- [x] PR does not match another non-stale PR currently opened

## Features

This solves a part of the open tasks from https://github.com/devicons/devicon/pull/1998#issuecomment-1807542194

* [x]  digitalocean (#2005)
* [x]  gatsby (#2006)
* [x]  mysql (#2007)
* [x]  nodejs (#2011)
* [x]  opensuse (#2002)
* [x]  postcss (#2012)
* [ ]  spring
* [ ]  typo3

## Notes

I will check on the opensuse icons later. the svg files currently include inkscape metadata. either there is a open PR or i add it later

PR #1998, opened 2023-11-12
Build preparation for release v2.16.0

Snailedlt:
```
A huge thanks to all our maintainers and contributors for making this release possible!

 **264 New Icons**
- [update icon: unity (original, original-wordmark, plain, plain-wordmark, line, line-wordmark)](https://github.com/devicons/devicon/pull/1994) by @lunatic-fox, @Snailedlt.
- [update icon: crystal (original, original-wordmark, line, line-wordmark)](https://github.com/devicons/devicon/pull/1992) by @canaleal, @lunatic-fox.
- [update icon: vercel (original, original-wordmark, line, line-wordmark)](https://github.com/devicons/devicon/pull/1991) by @canaleal, @lunatic-fox, @Snailedlt.
- [update icon: codepen (original, original-wordmark, line, line-wordmark)](https://github.com/devicons/devicon/pull/1990) by @canaleal, @lunatic-fox, @Snailedlt.
- [update icon: laravel (original, original-wordmark, line, line-wordmark)](https://github.com/devicons/devicon/pull/1985) by @canaleal, @lunatic-fox, @Snailedlt.
- [update icon: traefikproxy (original, original-wordmark, plain-wordmark, line, line-wordmark)](https://github.com/devicons/devicon/pull/1983) by @canaleal, @Snailedlt.
- [update icon: traefikmesh (original, original-wordmark, plain-wordmark, line, line-wordmark)](https://github.com/devicons/devicon/pull/1982) by @canaleal, @Snailedlt.
- [update icon: sonarqube (original, original-wordmark, plain-wordmark, line, line-wordmark)](https://github.com/devicons/devicon/pull/1981) by @canaleal, @Snailedlt.
- [update icon: msdos (original, plain, line)](https://github.com/devicons/devicon/pull/1980) by @canaleal, @lunatic-fox.
- [update icon: gimp (original, original-wordmark, plain, plain-wordmark, line, line-wordmark)](https://github.com/devicons/devicon/pull/1979) by @canaleal, @lunatic-fox.
- [new icon: neovim (original, original-wordmark, plain , plain-wordmark, line, line-wordmark)](https://github.com/devicons/devicon/pull/1962) by @Snailedlt.
- [update icon: aarch64 (original, plain, line)](https://github.com/devicons/devicon/pull/1961) by @Snailedlt.
- [update icon: yarn (original, original-wordmark, line, line-wordmark)](https://github.com/devicons/devicon/pull/1960) by @Snailedlt.
[...]

PR #2002, opened 2023-11-13
update icon: opensuse (original, original-wordmark)

weh:

## Features

- Replace plain versions with alias
- Optimize svg with vecta.io

**This PR is related to #2001 and https://github.com/devicons/devicon/pull/1998#issuecomment-1807542194**


canaleal (approved):

Snailedlt (approved):
image
image
image

Issue #1933, opened 2023-10-24
[FEATURE REQUEST] Migrate to trunk based development


Snailedlt:
### I have searched through the issues and didn't find my problem.

- [X] Confirm

### Problem

Currently we are using a workflow similar to GitFlow and have two major branches: `master` and `develop`.

This causes a lot of overhead when it comes to releases, and can make things overly complicated. 

### Possible Solution

A much simpler approach would be to just have a single `master` or `main` branch, and merge all PR's into that. Then once we have something shippable all we have to do is run a workflow to make the release.

### Additional information

In order to make the transition to trunk based development there's a large amount of automation scripts that need to be modified. The good thing is that there will also be quite a few that we can get rid of entirely. For example the [in-develop-labeler](https://github.com/devicons/devicon/blob/master/.github/workflows/in_develop_labeler.yml).
Going to a trunk based development model will drastically improve our agility, and reduce the project's complexity

Snailedlt:
I suggest we start on this right after the next release!

PR #2040, opened 2023-11-22
bugfix/fix: references of `faunadb` and `ceylon`

lunatic-fox:

## Features

**This PR partially closes #1998**

### DIFF
--- a/devicon.json
+++ b/devicon.json
@@ -1447,7 +1447,8 @@
                 "plain-wordmark"
             ],
             "font": [
-                "plain"
+                "plain",
+                "plain-wordmark"
             ]
         },
         "color": "#AB710A",
@@ -3034,7 +3035,9 @@
         "versions": {
             "svg": [
                 "original-wordmark",
-                "original"
+                "original",
+                "line",
+                "line-wordmark"
             ],
             "font": [
                 "original-wordmark",


Snailedlt (approved):
I'm not a 100% sure this will work for the build workflow, but let's merge it and see if it does ;)

If it doesn't we'll just have to create a separate PR for each icon

Snailedlt:
@lunatic-fox ahh, if that's the case then I'm pretty sure it won't work. Since it uses the PR title to know which icons it should add to the font and devicon.css. That's why the build failed last time when we changed the name of an icon, because it was looking at an old PR with the name `3dsmax`, but in the repo the `3dsmax` icons didn't exist... only `threedsmax`.
This is unfortunate... but I don't think we need to fix it since we'll have to change to a different way of building anyways once we move to trunk based development in #1933

lunatic-fox:
> @lunatic-fox ahh, if that's the case then I'm pretty sure it won't work. Since it uses the PR title to know which icons it should add to the font and devicon.css. That's why the build failed last time when we changed the name of an icon, because it was looking at an old PR with the name `3dsmax`, but in the repo the `3dsmax` icons didn't exist... only `threedsmax`.
> This is unfortunate... but I don't think we need to fix it since we'll have to change to a different way of building anyways once we move to trunk based development in #1933 

Hmm... so, can you please revert this PR, I'll split it when I'm back home at night.

PR #2048, opened 2023-11-23
update icon: typo3 (original, original-wordmark, plain-wordmark, line, line-wordmark)

weh:

## Features

- Removed padding and optimized images
- Replaced duplicate original/plain with alias
- Added line versions

**This PR closes NONE**

Part of #1998 and #2001


lunatic-fox (dismissed):
Good job! 👍🏼
image
image
image
image

Snailedlt (commented):
The plain and line versions are missing the fill attribute, which makes those versions black. That's fine for the font since we can change the color easily, but for the svg versions it's more difficult to change the color.
imo we should use the color from `devicon.json` as fill for all icons with a single path

@lunatic-fox I know you have some opinions on this, what do you think?

lunatic-fox:
> The plain and line versions are missing the fill attribute, which makes those versions black. That's fine for the font since we can change the color easily, but for the svg versions it's more difficult to change the color. imo we should use the color from `devicon.json` as fill for all icons with a single path
> 
> @lunatic-fox I know you have some opinions on this, what do you think?

I got your point @Snailedlt and yeah, all this time I was considering `devicon.json` color just for font versions, but didn't consider for SVG versions. Making changes on the next commit.

lunatic-fox (approved):
image
image
image
image

Snailedlt (approved):
LGTM! ✔️ 
Thank you for another great contribution! 💯 🙇

PR #2025, opened 2023-11-20
new icon: fiber (original, plain, line)

canaleal:
## Double check these details before you open a PR

- [x] PR does not match another non-stale PR currently opened
- [x] PR name matches the format *new icon: <i>Icon name</i> (<i>versions separated by comma</i>)*. More details [here](https://github.com/devicons/devicon/blob/develop/CONTRIBUTING.md#overview)
- [x] PR's base is the `develop` branch.
- [x] Your icons are inside a folder as seen [here](https://github.com/devicons/devicon/blob/develop/CONTRIBUTING.md#organizational-guidelines)
- [x] SVG matches the standards laid out [here](https://github.com/devicons/devicon/blob/develop/CONTRIBUTING.md#svgStandards)
- [x] A new object is added in the `devicon.json` file as seen [here](https://github.com/devicons/devicon/blob/develop/CONTRIBUTING.md#-updating-the-deviconjson-)

**This PR closes #1971**

## Link to prove your SVG is correct and up-to-date.

- [Github](https://github.com/gofiber/fiber)
- [Fiber SVG](https://gofiber.io/assets/images/logo.svg)
- [Github Branding](https://github.com/gofiber/fiber/blob/master/.github/testdata/fs/img/fiber.png)

The original, plain, and line svgs can be used for the wordmarks using aliases. 
The color `#00acd7` was selected for the font SVGs as it's the main Go Fiber color.

PR #2259, opened 2024-09-03
New icon fiber patch

Ryrden:
Trying to solve the problem from PR #2025


Ryrden:
I made a mistake probably here, ignore it

PR #2261, opened 2024-09-05
update icon: angularjs (plain)

Finii:
[why]
There are two problems with the plain svg:
- The top of the 'A' consists of a lot of nodes that have a very complex outline, while in reality it should only be one sharp node
- The 'A' outline is not closed

[how]
Remove superfluous nodes from outline and close it.

## Features

General location of the problems:

image

Top of the 'A' shape, these are 10 nodes instead of 1 :grimacing: 

image

Nodes 1 and 2 are not connected, note the missing direction arrow

image

**This PR closes NONE**

## Notes

The non-closed outline breaks direct import into fontforge.

I did not check the other 3 angularjs icons, most likely they are there also.


Snailedlt (dismissed):

Snailedlt (approved):
@Finii The changes look good, but they should be merged into the `develop` branch instead of `master`. I changed it for you though, so all good now :)

canaleal (approved):

PR #2263, opened 2024-09-05
update icon: bash (original, plain)

Finii:
[why]
When calculating the bounding box of the icon the real icon is too far right. There is in fact a invisible box on the real icon's left.

[how]
Edit `svg` file directly (in vi) and just remove the outline's description.

## Features

Load icon into `inkscape` and just select all - note the ghost square on the left ?!

image

Removed path marked blue:

image

**This PR closes NONE**

## Notes


canaleal (approved):
Great work. 

Hope you don't mind but I also updated the original svg as the old version did not reach the edges of the viewbox. 

image

canaleal (approved):

PR #2264, opened 2024-09-05
update icon: llvm (plain, line)

Finii:
[why]
While the plain icon is plainer than the original icon, it is still quite complex with very thin strokes.

[how]
Just take the outermost outlines of the plain icon and crate a solid icon from it.

### Note

This can be **controversial**.
For me the plain icon is not really plain, it has lots and lots of detail and consists of 2600 nodes:

image

The simple plain icon proposed here drops all the details and uses just the outline (260 nodes)

image

If one needs an icon with all the glory details the `-original` should be a good fit, the plain should be a simple(r) version, I believe. But anyhow, this is very opinionated and I do not want to mess with your ideas here. For a small text icon the current plain icon is far too complex and all the small detail hairlines can not be seen anyhow.

## Features

**This PR closes NONE**

## Notes


Finii:
Hmm, hand-remove the gaps in the tail, that is an obvious artifact of the fill related broken outline in the original svg.
Force push.

image

Snailedlt (commented):
hmmm, I don't agree with this one. While I agree that the icon could benefit from some simplification I think going all black is too much simplification, making it less recognizable... but let's see what the other maintainers think too :)

### Edit: 
Good that you fixed the artifact though. That should probably be fixed in the original icon too.

Some context: The original icon is made by converting a png to svg with inkscape. The process is not ideal so there are a few artifacts. I think it could be improved upon now though, since I know there are better png to svg converters out there now than when we made this icon :)

Finii:
> hmmm, I don't agree with this one

:laughing: 

As I said, controversial (see image below).

For me, as "in a font" the current plain one is just too intricate. Have a look at it sized down:

image

But that is just for ME and font usage, which might or might not be important for you.
Just wanted to share this icon version with you.

image

Snailedlt:
I agree it looks better in the font, but in larger sizes it looks significantly worse. Perhaps we could find some middle ground?
One way would be to roughly inverse the current plain version, and remove some of the detailing lines.

I'm thinking something like this:
image

Snailedlt:
@Finii
Here's a suggestion. Though it should be optimized more, because it's very large (80Kb)
llvm-original

Snailedlt:
Sidenote: The old `plain` version can still be used as a `line` version instead :)

weh (dismissed):
great work :+1:

Snailedlt (changes requested):
I don't believe the suggested version is close enough to the original to justify the original. Most of the lines should be kept imo. I added a suggestion as to how it could look in a previous comment, I think if some of the nodes are removed on that version we'll have both a pretty nice looking and optimized plain version

Finii:
Already in this very fast threshold-to-black-and-white image one can glimpse that the outlying lines are thinner (and should than vanish by combining with the inner areas) while the inside lines really appear more fat more substantial and would be converted to white/transparent lines in a monochrome version.

:thinking: 

image

Finii:
The design problem reminds me of that one spiky fish icon, that I solved in a comparable manner, lets see if I can find it...

Ah, it was OpenBSD, here: https://github.com/lukas-w/font-logos/pull/93, but apparently did not materialize?

image

image

Finii:
Something like this, this is a rough sketch needing cleanup:

image

Snailedlt:
I'm personally very happy with the last suggestion, but I haven't looked very closely at the result, so let me know if you find any flaws

Finii:
Your plain version looks awesome!

Zooming in (randomly, on the right paw (left in the image)) is this which shows some technical problems

image

There is this blob of 3 nodes that is 'dirt', and then the arrowed nodes are in fact two nodes.

I set fill to none and turned on 0.2 px stroke to see defects in the path. Did not notice any more 'smudges', but there are a lot of two-node-paires with the exact same coordinates. Inkscape shows them smaller:

image

I encountered these strange pairs in the past and my belief is that the path is not 'really' closed here, but this is start and end on the same place, but not in the same node.

Not really a clue if they are generally a problem, but sometimes they are. This is not clear for me.
Anyhow, breaking apart the path and combining it again removes some but not all of the double points. :thinking: 

So my intermediate conclusion is that your new plain icon is great; but you can remove that three-node-smudge between its right hand fingers (left side in the image).
I will investigate a bit more into the two point problem if time permits. Maybe tools like svgo already attack them.
Inscape simplify does not in general. Also simplify does not really enhance your plain svg, so that seems to be already quite compact :+1:

Finii:
Well, my line version was just quickly done in boring zoom meetings :grimacing: 
I can improve it taking your comments into account, but not today but I assume tomorrow.

Snailedlt:
Sounds good to me. Let's try to get this PR merged before the end of next week :)

Finii:
Use @Snailedlt's plain version sans the smudge at the right hand (which I called left hand in the commit message, sigh).

Use @Snailedlt's line version and increase the outer line a bit to make it as thick as the inside lines.

Force push to
* remove the unneeded merge commits
* have just the final changes and not the development in the repo

devicon.json
```diff
@@ -7047,7 +7047,8 @@
         "versions": {
             "svg": [
                 "original",
-                "plain"
+                "plain",
+                "line"
             ],
             "font": [
                 "plain"
```
Snailedlt, line 7054: Add line icon to font
```suggestion
                "plain",
                "line"
```

Snailedlt (changes requested):
Line version needs to be added to the font. Let's also add the color as fill to the line and plain versions :)
`#5A90B6`

devicon.json
Finii: Applied

Snailedlt:
It's more of a convention we've used since it makes it easier for some third-party tools. It's not really needed for our internal workflow.
Yeah, maybe we should update the wiki.

I don't even remember why we re-added the fill for many icons at this point 😅

icons/llvm/llvm-line.svg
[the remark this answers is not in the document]
Finii: I'll change both (i.e. add concrete fill color and fix the attributes), force push imminent.

I wonder how the attributes got there as I think (!) I explicitly disabled them in the Inkscape export.

Finii:
Dear SVG Checker Bot,

it would probably be more helpful if you hint how to sort the `devicons.json` file.
For example users could just

```shell
$ jq --slurp --indent 4 '.[] | sort_by(.name)' < devicon.json  > devicon_sorted.json
$ mv devicon_sorted.json devicon.json
```

I will add a commit with the result of this. Lets see if you are content then.

ReenigneArcher (approved):
The order of the json file shouldn't hurt anything, and since everything is supposed to be alphabetical I think this is a good change.

In the future I want to add a PR check that makes sure the json file is properly formatted and properly sorted which can run for every PR.

For the icons, any idea how to fix this? The viewbox is the right size, but it displays as the wrong size.

image

Approving since both of the things I discussed would be future improvements.

Snailedlt:
@ReenigneArcher yeah, adding a check for formatting and sorting would be nice!

I'm not sure why you have that sizing issue, in my GitHub preview it looks fine: 
image

Snailedlt:
@canaleal I know you already started the release process, but do you think we can merge this as well, since there were some issues with the font?

PR #2290, opened 2024-10-25
new icon: delphi-12 (original)

Bulani:

**This PR closes NONE**

https://www.embarcadero.com/br/news/logo


canaleal:
Excellent work on the icon. If possible could you also create a plain version of the icon? Here is how a [plain](https://github.com/devicons/devicon/wiki/SVG-Versions) icon should look like.

Bulani:
> Excellent work on the icon. If possible could you also create a plain version of the icon? Here is how a [plain](https://github.com/devicons/devicon/wiki/SVG-Versions) icon should look like.

Added logo in plain version

Bulani:
@canaleal Can you review this PR, now everything is ok

canaleal:
@Snailedlt Opinions on the original.svg? It had a lot of paths and I don't think there's a way to optimize it.

Snailedlt:
Hmm, if you're not able to optimize it I doubt I'll be able to do it.

Maybe the guy from nerdfonts would be able to help here? I know he is good at optimizing the number of nodes.

@finii do you have any tips for us to optimize this icon? Would be nice if it was properly optimized here, so you don't have to re-optimize it for usage in nerdfonts

Finii:
Hej ;)

> It had a lot of paths and I don't think there's a way to optimize it.

In Inkscape, select all, `Path -> Exclusion` combines all path into one. That is not always the way, but here it certainly is. You end up with one path.

Removed this tiny (probably unwanted) path first:

image

And then the fill was concrete colors for each path, different colors :thinking: 

image

In the simple version you probably want no concrete color at all, select 'filled with unknown color' and probably 'evenodd fill', and 'no stroke'. Fonts always work with even-odd fill, and when you save it here with nonzero fill that often is converted correctly but not always.

image


And then I guess there are too many nodes, optimized them away. Here the `Path -> Simplify` consideres its accuracy based on all selected nodes area (lengths etc) (and only simplifies the selected nodes) relatively; so if you want to make smallest simplifications you can select just some few nodes, while selecting all nodes in one allows bigger positional errors in total.

Here I simplified all nodes except the outer circle which is already optimal with 4 nodes at the extrema.

image

_Left side 131 nodes, right side 347 nodes_

And finally ... `Save As...`, use 'Optimized svg`. Let it clean up everything and usually 4 digits is enough to avoid rounding errors.

```console
$ ll delphi-plain*         
-rw-rw-r-- 1 fini fini 10770 Feb 10 10:46 delphi-plain_orig.svg
-rw-rw-r-- 1 fini fini  4906 Feb 10 11:08 delphi-plain.svg
```

Here's the file:

[delphi-plain](https://github.com/user-attachments/assets/227587b5-3c57-4d65-8578-da25e62242dd)

Snailedlt:
@Finii thanks a bunch! I'm glad we have such a knowledgeable developer we can reach out to for help!
I assume the original version can also be simplified by using some of these tricks. I'm guessing mainly by removing the unused nodes and bringing down the nodecount. Let me know if thee are any other things we could do to optimize the delphi-original.svg specifically.

@canaleal I think these should be good, just make sure you double check that the plain versions render correctly on icomoon :)

Finii:
> have such a knowledgeable developer

Hobbyist :grimacing: 

> I assume the original version can also be simplified by using some of these tricks.

image

2000 path with maybe 300 nodes each :grimacing: 

The reason is that the fill is with gradients, but the gradients are in fact just a lot of individual paths all filled with one solid color (red), slightly different from the path nearby.

image

_Top of the icon zoomed way in_

Note all the single colored areas, and how 'unclean' the darkred wedge comes in :grimacing: 

But here I have no clue how to optimize it while keeping the coloring true to the original. A 'regular' gradient is how it is and can not be tweaked with infinite precision. One would need to define which 'imperfections' or 'deviations from the original' or still ok.

:thinking: I am not even sure what the aim of the different svgs here is, maybe to be consumed as svgs, and as such that is probably all good. Otoh, the more I look at icons, the more I wonder WHO create the original svgs, probably graphic designers that do not care and/or understand how that all 'really' works. Like me 1992 with Corel Draw 2 :rofl:

Snailedlt:
Yeah, it was probably made with some workarounds in a tool that wasn't made for gradients I guess.

Perhaps we should ask the delphi developers if we can remake the icon with a simplified gradient. I bet they would also appreciate a more optimized version

gewaleelek:
So I tried to just manually recreating everything. The gradients aren't exactly the same, of course, but maybe close enough?

delphi-original

Snailedlt:
@gewaleelek that looks very good in my opinion, but I think we should contact the delphi team to hear what they think. Maybe they have a very clever way to optimize the icon already :)

@sgratzl perhaps you have some input here?
We want to add the delphi logo to our icon library, but we're having trouble optimizing the icon. Do you or anyone else in the Delphi team have any opinions about how it can be optimized for usage in fonts and SVGs on websites? @gewaleelek made an optimized version from scratch that might interest you :)

Bulani:
any updates?

ReenigneArcher:
Should this be made more clear that this not Delphi, the company? https://www.delphiautoparts.com/

Snailedlt:
> any updates?

Afraid not

gewaleelek:
> @gewaleelek that looks very good in my opinion, but I think we should contact the delphi team to hear what they think. Maybe they have a very clever way to optimize the icon already :)
> 
> @sgratzl perhaps you have some input here? We want to add the delphi logo to our icon library, but we're having trouble optimizing the icon. Do you or anyone else in the Delphi team have any opinions about how it can be optimized for usage in fonts and SVGs on websites? @gewaleelek made an optimized version from scratch that might interest you :)

I'm just wondering but is you tagged involved with Delphi (the IDE)? Their profile mentions them being in the "Delphi" group, but _that Delphi_ seems to be related to [healthcare data for infectious diseases](https://delphi.cmu.edu/), not the commercial Object Pascal IDE.

Snailedlt:
@gewaleelek ohh, it seems your right.
Do you know of anyone we could tag in the correct delphi team?

gewaleelek:
Unfortunately no. This PR is the first time I've heard of this tool, actually. There's a [contact email](https://www.embarcadero.com/company/contact-us) on their website, at least.

In my humble opinion, with the officially provided "vector" looking like it was just converted from a raster image instead of the other way around, I can't help but wonder if their response will be very helpful at all.

Bulani:
> Should this be made more clear that this not Delphi, the company? https://www.delphiautoparts.com/

Official website: https://www.embarcadero.com/

And i got the logo in: https://www.embarcadero.com/news/logo

Snailedlt:
@Bulani did you make the original svg from the PNG or EPS version on the website?

Snailedlt:
> > @Bulani did you make the original svg from the PNG or EPS version on the website?
> 
> @Snailedlt I got the .epg and convert to .svg, and a used the guides:
> 
> https://github.com/devicons/devicon/wiki/Creating-original-Icons
> 
> https://github.com/devicons/devicon/wiki/Optimizing-icons

@Bulani okey, great. That's the right way to do it :)

I was just trying to understand if the color banding on the gradients was a result of the conversion to SVG or not, but if it was converted from eps I think the color banding is also in the original icon.

The original icon is over 1.2 million characters long 😅

I suggest we use the optimized version created by @gewaleelek so that the size of the icon is manageable.

Snailedlt:
whoopsie. Seems I did not check the plain icon well enough.

Reverted. I'll create a new PR

PR #2270, opened 2024-09-15
feat(languages): use github language names and add new icons

ReenigneArcher:
## Double check these details before you open a PR

- [x] PR does not match another non-stale PR currently opened

## Features

This PR adds language icons and alias that match the naming GitHub uses in the GitHub API. GitHub uses "linguist" for the languages API. Their list of languages can be found here: https://github.com/github-linguist/linguist/blob/master/lib/linguist/languages.yml

The advantage of this is it makes it possible to use icons directly, without creating a custom mapping.

- Many new icons, from https://github.com/LizardByte/uno/tree/master/gh-pages-template/language-icons
- powershell icon updated
- current nginx icon moved to wordmark icon and added new original icon
- jest-plain renamed to jest-original
- fixed graphql "plain" icons -> original
- re-ordered a couple of out of order icons, like windows8 and reactrouter
- some new altnames added (taken from the linguist list)
- some tags added or changed from existing icons (taken from the linguist list)

**This PR closes NONE**

- Resolves https://github.com/devicons/devicon/discussions/2266

## Notes

I understand that typically new icons are probably submitted one per PR; however since this is somewhat of a new feature (names to match github API), I'd really rather avoid 230 PRs.

The colors for new icons were selected using a color picker. If there was a more prominent color, I selected that, otherwise it may be a random color in some cases (where an icon may have many colors).

I have a question on "graphql". There is no "original" icon, and no alias to it. Is that correct? Edit: I updated the graphql icons based on the comment below.

PR #2295, opened 2024-11-05
ci: allow multi submission

ReenigneArcher:

## Features

This PR allows multi submission edits in a single PR, as discussed in https://github.com/devicons/devicon/pull/2270.

**This PR closes NONE**

## Notes

I have tested this against the same branch that creates https://github.com/devicons/devicon/pull/2270 in my local fork. This is the resulting comment: https://github.com/ReenigneArcher/devicon/pull/1#issuecomment-2456068176. All of the comments are valid issues, so I believe this is working as expected.

Note, that somewhere in the code empty strings are getting added to the error list. I don't believe this is anything from my changes, but I have just removed the empties using the following line.

`err_msg = list(filter(None, err_msg)) # remove empty strings from err_msg`

Some trailing whitespace was also automatically removed from lines in modified files.

@Snailedlt FYI

### DIFF
--- a/.github/scripts/build_assets/arg_getters.py
+++ b/.github/scripts/build_assets/arg_getters.py
@@ -2,9 +2,9 @@ from argparse import ArgumentParser
 from build_assets.PathResolverAction import PathResolverAction
 
 
-def get_selenium_runner_args(has_token=True, peek_mode=False):
+def get_selenium_runner_args(has_token=True):
     """
-    Get the commandline arguments for the icomoon_peek.py and 
+    Get the commandline arguments for the icomoon_peek.py and
     icomoon_build.py.
     """
     parser = ArgumentParser(description="Upload svgs to Icomoon to create icon files.")
@@ -33,12 +33,13 @@ def get_selenium_runner_args(has_token=True, peek_mode=False):
                         help="The download destination of the Icomoon files",
                         action=PathResolverAction)
 
-    if peek_mode:
-        parser.add_argument("pr_title",
-                            help="The title of the PR that we are peeking at")
     if has_token != False:
         parser.add_argument("token",
                             help="The GitHub token to access the GitHub REST API.")
+
+    parser.add_argument("changed_files",
+                        help="List of SVG files changed since the last release/tag",
+                        nargs="+")
 
     return parser.parse_args()
 
@@ -49,9 +50,6 @@ def get_check_icon_pr_args():
     """
     parser = ArgumentParser(description="Check the SVGs to ensure their attributes are correct. Run whenever a PR is opened")
 
-    parser.add_argument("pr_title",
-                        help="The title of the PR that we are peeking at")
-
     parser.add_argument("icons_folder_path",
                         help="The path to the icons folder",
                         action=PathResolverAction)
@@ -59,6 +57,10 @@ def get_check_icon_pr_args():
     parser.add_argument("devicon_json_path",
                         help="The path to the devicon.json",
                         action=PathResolverAction)
+
+    parser.add_argument("changed_files",
+                        help="List of SVG files changed in the PR",
+                        nargs="+")
 
     return parser.parse_args()
 
--- a/.github/scripts/build_assets/util.py
+++ b/.github/scripts/build_assets/util.py
@@ -50,29 +50,26 @@ def set_env_var(key: str, value: str, delimiter: str='~'):
         raise Exception("This function doesn't support this platform: " + platform.system())
 
 
-def find_object_added_in_pr(icons: List[dict], pr_title: str):
+def find_changed_icons(icons: List[dict], changed_files: List[str]) -> List[dict]:
     """
-    Find the icon name from the PR title. 
+    Find the changed icons provided in the changed_files list.
     :param icons, a list of the font objects found in the devicon.json.
-    :pr_title, the title of the PR that this workflow was called on.
-    :return a dictionary with the "name"
-    entry's value matching the name in the pr_title.
-    :raise If no object can be found, raise an Exception.
+    :param changed_files, SVG files changed in the PR or since the last release/tag.
+    :return a list of dictionaries with the "name"
+    entry values matching the name of changed icons.
     """
-    try:
-        pattern = re.compile(r"(?<=^new icon: )\w+ (?=\(.+\))|(?<=^update icon: )\w+ (?=\(.+\))", re.I)
-        icon_name_index = 0
-        icon_name = pattern.findall(pr_title)[icon_name_index].lower().strip()  # should only have one match
-        icon = [icon for icon in icons if icon["name"] == icon_name][0]
-        return icon
-    except IndexError as e:  # there are no match in the findall()
-        print(e)
-        message = "util.find_object_added_in_pr: Couldn't find an icon matching the name in the PR title.\n" \
-            f"PR title is: '{pr_title}'"
-        raise Exception(message)
+    filtered_icons = []
+    icon_names = []
+    for file in changed_files:
+        icon_name = Path(file).parent.name
+        icon = [icon for icon in icons if icon["name"] == icon_name]
+        if len(icon) > 0 and icon_name not in icon_names:
+            icon_names.append(icon_name)
+            filtered_icons.extend(icon)
+    return filtered_icons
 
 
-def is_svg_in_font_attribute(svg_file_path: Path, devicon_object: dict): 
+def is_svg_in_font_attribute(svg_file_path: Path, devicon_object: dict):
     """
     Check if svg is in devicon.json's font attribute.
     :param svg_file_path, the path to a single svg icon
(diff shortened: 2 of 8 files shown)


Snailedlt:
Great!
I'll see if I can take a look at the PR this weekend

Snailedlt (changes requested):
This looks very good!
However I think we might get issues with the release script, since that also builds icons based on the PR title.
I might be wrong here, but iirc that's a different script, which will also need to be refactored. 

To test out if this works with the release script(s), you can go through [the release guide](https://github.com/devicons/devicon/wiki/Release-strategy,-conventions,-preparation-and-execution#execution) on your fork, and see if you get any errors, and if every step executes as expected.

If you find any ways to make the release process simpler, please feel free to adjust it as you see fit :)

ReenigneArcher:
> However I think we might get issues with the release script, since that also builds icons based on the PR title.

Thanks for the review. I'll dig into that in the next few days and report back.

ReenigneArcher:
Just pasting this code snip here so I can easily find it when I go to finish this:

https://github.com/devicons/devicon/blob/ca28c779441053191ff11710fe24a9e6c23690d6/.github/scripts/icomoon_build.py#L72-101

I actually believe this may already work as is due to this section:

```python
    # get any icons that might not have been found by the API
    # sometimes happen due to the PR being opened before the latest build release
    new_icons_from_devicon_json = filehandler.find_new_icons_in_devicon_json(
        devicon_json_path, icomoon_json_path)
```

ReenigneArcher:
@Snailedlt I'm not sure I can complete a release test in my fork.

For some reason the `build_icons.yml` workflow is not appearing under the actions tab of my repo.

https://github.com/ReenigneArcher/devicon/actions
image

I've only ever seen this when the workflow file does not exist on the default branch, but that is not the case here. Any idea?

Snailedlt:
@ReenigneArcher hmmm, that's very odd. No idea sadly... but maybe it helps to recreate the repo instead of forking it? You could also try re-forking it

ReenigneArcher:
I re-forked the repo into another org, and that one seems to be fine. I opened a ticket with GitHub support, hopefully they can solve it as I'd prefer not to delete my current fork.

ReenigneArcher:
@Snailedlt I got it to appear by making a small adjustment to the workflow (no functional change)... but the workflow fails due to deprecated actions. I will make a PR updating the deprecated actions, and I can also introduce dependabot which can automatically keep workflow actions up to date by creating PRs when new versions are released. Here's an example PR: https://github.com/LizardByte/Sunshine/pull/3399

ReenigneArcher:
@Snailedlt could you take a look at #2310? This PR depends upon that one, and actually any future workflow runs in this repo as many of these outdated actions are going to stop working altogether (as they are automatically failing workflows): https://github.blog/changelog/2024-02-13-deprecation-notice-v1-and-v2-of-the-artifact-actions/

Snailedlt:
@ReenigneArcher Now that #2310 has been merged, this PR should be possible to test, right?

ReenigneArcher:
@Snailedlt yes, I believe so. I just need to re-gather my thoughts on this as it has been a a while since I looked at all this.

ReenigneArcher:
I started looking at this again tonight. I found some things I missed which is bring up some questions.

1. This will affect the `icomoon_peek.py` file as well. I checked the action logs and this workflow is ALWAYS skipped. https://github.com/devicons/devicon/actions/workflows/peek_icons.yml ... Can I remove the workflow and the script? I'm not sure the history on this, but looks like this workflow is intended to run when the `bot:peek` label is added which hasn't been used since May 2023. https://github.com/devicons/devicon/pulls?q=is%3Apr+label%3Abot%3Apeek+is%3Aclosed
2. For some reason I am getting an exit code of 1, with nothing printed. To me this indicates it is even failing before this line: https://github.com/ReenigneArcher/devicon/blob/9436f2555dfe3e1626758394786789a05f507927/.github/scripts/icomoon_build.py#L27 ... Possibly in the args getter. I've used `argparse` quite a bit, but always with named arguments, never positional. Normally when argparse fails, it's prints out a "usage" statement, so I'm still not sure if that's it. Any ideas?

I have a commit in the testing branch which doesn't exit in my PR (I will bring everything in once I get it working): https://github.com/ReenigneArcher/devicon/commits/draft-release/

Snailedlt:
1. The bot peek script stopped working a long time ago. I believe because of an PAI change, or some bandwidth limit in the API. When it worked it would post a few screenshots of how the icon font looks when uploaded to icomoon. It would be awesome if we got it working again, since it's a hazzle to upload and check the font manually... But it's not required to get this PR merged. So feel free to try and fix it, or delete it. Whichever you want :)
2. I've got no idea sadly. Logging the arguments might be a good place to start though. I think you're right about it exiting before that line, maybe even before that method is called. Maybe you can try changing to named arguments, since that makes it quite a bit easier to read.

ReenigneArcher:
1. With that information, I will keep the script and adjust to accept the potentially multiple changed icons in a PR. I'll save trying to get the script actually working again for a later PR though.
2. I may make a separate PR to start using named arguments, just so this PR doesn't become too overwhelming to review.

Snailedlt:
> I figured out why there was no output. It seems the output is redirected to a log file which is then uploaded as an artifact.

Right, iirc it's uploaded as an artifact, and then later on the artifact is retrieved. It's a weird workaround for a very odd security issue with allowing workflows to run on forked repos. 
Here's an explanation of why we need to upload the artifact in a preflight script, and then download it again to use it: https://github.com/devicons/devicon/issues/1099

This is also related: https://github.com/devicons/devicon/issues/1715


> KeyError: "There is no item named 'fonts/devicon.eot' in the archive"

Hmm, seems there is something wrong with the font creation in icomoon. devicon.eot is a file that should be downloaded from icomoon when creating the font.
Maybe you could try creating the font manually. 


I also recommend reading about how we build the font in our wiki to get a better grasp of how it all works. It's been a while since I read it myself, so maybe I should too 😅
Anyways, here are some links for you: 
- https://github.com/devicons/devicon/wiki/Release-strategy,-conventions,-preparation-and-execution
- https://github.com/devicons/devicon/wiki/How-We-Create-The-Icons
- https://github.com/devicons/devicon/tree/develop?tab=readme-ov-file#build-the-new-icons

This wiki page is also very useful for an overview on most of our automated workflows, and it should also be updated when we update this one: https://github.com/devicons/devicon/wiki/Our-automated-tasks-and-bots

Snailedlt:
Hmm, it might also be that icomoon has renamed the .eot file to something else, or that the GitHub cache is stuck or something.
I would suggest trying these two things first:
1. Delete the cache for the workflow, and rerun it to see if that fixes anything
2. Do the font creation process manually, to better understand how it works, and to see if the filenames and locations are as expected.

ReenigneArcher:
> Hmm, it might also be that icomoon has renamed the .eot file

This was my thought as well, or that it was removed altogether. I'm not too versed in font creation.

I don't have any caches in my fork, so I don't think that is the issue. I will start with printing out all the contents of the archive, and reading the above references.

Snailedlt:
Sounds like a good plan. Let me know if you get stuck :)

ReenigneArcher:
Here are my findings so far:

1. The file is not in the downloaded zip. These are the contents: `['selection.json', 'fonts/', 'fonts/devicon.ttf', 'fonts/devicon.woff', 'fonts/devicon.svg', 'demo.html', 'demo-files/', 'demo-files/demo.js', 'demo-files/demo.css', 'Read Me.txt', 'style.css']`

I manually tried to create a font via the icomoon website (only with one of the default icons), downloaded the archive and it did contain the `.eot` file.

image

I also edited the workflow to upload the zip as an artifact:
image

My first thought was the archive was not fully downloaded; however I doubled the wait delay and the result was identical so this theory is probably not correct.

I'm not sure what else to check/try on this.

Snailedlt:
Hmm, that's interesting 🤔 

So I'm guessing the issue is somewhere in the [build_icons.yml](https://github.com/devicons/devicon/actions/workflows/build_icons.yml) script then. Either that or icomoon changes their API.

Have you checked if the .eot file is created when doing it on the master branch?

If it is, then there must be something we changed after the last release that broke it

ReenigneArcher:
Unfortunately, the workflows in master do not work due to the EOL GitHub actions. But:

- Nothing in scripts could be causing this as they haven't changed since the last release (other than a non breaking update to `requests`): https://github.com/devicons/devicon/tree/develop/.github/scripts
- Looking at the workflow changes, I don't see any changes since the last release either (other than action updates which wouldn't cause this)

ReenigneArcher:
I found the issue with the missing `.eot` file. The fix is in https://github.com/devicons/devicon/pull/2361

ReenigneArcher:
Closing because I think this would make the review process too complex.

PR #2360, opened 2025-02-15
build(npm): automatically run build-icons for correct os

ReenigneArcher:

## Features

This PR makes it so `npm run build-icons` works on all standard OSes.

**This PR closes NONE**

## Notes

A new dev dependency was added to simplify this: https://www.npmjs.com/package/run-script-os

Edit: This is draft as I discovered it is not using the activated python venv. Will update once resolved.

### DIFF
--- a/README.md
+++ b/README.md
@@ -269,10 +269,10 @@ https://www.python.org/downloads/
 > **Note**
 > Make sure your Python install includes [pip](https://pypi.org/project/pip/)
 
-<h3>Install Selenium</h3>
+<h3>Install Dependencies</h3>
 
 ```bash
-python3 -m pip install --upgrade pip && pip install selenium==4.1.0 requests==2.25.1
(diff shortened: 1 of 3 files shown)


ReenigneArcher:
I opened a bug report in npm for the venv issue. I think this would technically affect the unix commands as well so it's probably not necessary to wait for this to be resolved.

https://github.com/npm/cli/issues/8111

But, I do have an idea on how to detect if we are running in an activated venv which I will work on shortly.

ReenigneArcher:
This is working now, but discovered build failures due to an icon name mismatch (https://github.com/devicons/devicon/pull/2407).

ReenigneArcher:
This should be good to merge if anyone can review. Thanks!

canaleal (approved):

Snailedlt (approved):

Snailedlt:
Great work, thnaks for letting us know when the PR was done :)

Merged

PR #2361, opened 2025-02-15
build(icomoon): fix missing eot file and add font metadata

ReenigneArcher:

## Features

This PR fixes the issue of the missing `.eot` file when building the fonts.

The issue was that the `Support IE 8` checkbox was unselected in the `icomoon.json`. I also added some metadata to the `icomoon.json`

image
image

**This PR closes NONE**


ReenigneArcher:
As an alternative to this, we could programmatically modify the `icomoon.json` before the import process begins. Let me know if that's preferred over this approach.

Snailedlt:
@ReenigneArcher 
I don't have any preferences. Choose the one that you think is best.
Basically make sure it's as stable as possible while keeping it easily maintianable :)

ReenigneArcher:
@Snailedlt sounds good. I am good with this PR as is. It's the easiest and fastest approach to getting the build working again.

I know it's not easy to review this one since the json is minified.

The process I took was simply:

1. import the existing file to the icomoon app
2. adjust the settings
3. download the font
4. extract the json file from the archive and rename it to icomoon.json

ReenigneArcher:
What is needed to get this merged? Builds won't work until this is in.

canaleal (approved):

canaleal:
@ReenigneArcher All is good, just update this branch and it can be merged. 
Great work as always.

ReenigneArcher:
@canaleal I rebased the branch. Thank you!

PR #2011, opened 2023-11-16
update icon: nodejs (original, original-wordmark, plain, plain-wordmark, line, line-wordmark)

weh:
## Double check these details before you open a PR

- [x] PR does not match another non-stale PR currently opened

## Features

- Replace plain version with alias
- Remove padding
- Add newlines

**This PR closes NONE**

Part of #2001

PR #2418, opened 2025-05-07
update icon: nodejs (plain, plain-wordmark, line, line-wordmark)

Finii:
**[why]**
The plain/line item contains two lines that are not visible in the original icon. Technically they are there but they are hidden. These lines make it impossible to recognize the stylized N that is the basis of the icon.

**[how]**
Take the rounded hexagon from the `eps` file (because it was the most clean one) and hand draw the lines of the `N`. Create from that simple form all the new plain/line icons.

Use a different stroke thickness for the wordmark icons because the previously used line width is ridiculously thin.

### Images

Left: current, right: suggested new form

#### plain

image

#### line

image

#### plain wordmark

image

#### line wordmark

image

#### original for reference

image

This has been suggested by @Snailedlt here: https://github.com/devicons/devicon/pull/2011#issuecomment-2853428320

(Hope I got all the names, branches, and bits and pieces right this time :grimacing: )


Finii:
> I'm the `check-bot` and we have some issues with your PR:
> 
> ```
> devicon.json is not sorted correctly.
> Please make sure that your icon is added in the `devicon.json` file at the correct alphabetic position
> as seen here: https://github.com/devicons/devicon/wiki/Updating-%60devicon.json%60
> ```

Dear SVG-Checker Bot, I did not even touch that file, and it is exactly sorted as it is in `master`. As I added no new icons that means that either your checking algorithm is wrong, or that some maintainer pulled an 'invalid' `devicon.json` into master.
Anyhow, I can do nothing about anything here, I'm sorry.

Snailedlt:
Please ignore the check bot. We have a PR that should fix it, but for now it is kinda broken :/

Snailedlt (approved):
Great work @Finii !
Thank you for suggesting and fixing this issue! ❤️

The only thing I would suggest to improve the icons is to move the fill up before the path, so that it is a bit easier to see the diff and maintain it, but it's not a big enough deal to warrant non-approval.

Great work as always! ✅

Finii:
> The only thing I would suggest to improve the icons is to move the fill up before the path, so that it is a bit easier to see the diff and maintain it, but it's not a big enough deal to warrant non-approval.

Whatever, that is simple. Just changed the files in an editor and re-used the 'preamble' of the svgs in the repo instead of 'stock inkscape' (see commit message).

:+1:

Snailedlt (commented):
LGTM! ✔️

Snailedlt (approved):

PR #2427, opened 2025-06-02
Revert "new icon: delphi (original, plain)"

Snailedlt:
Reverts devicons/devicon#2290


Finii:
Hmm, in fact it has 3 colors + transparent

image

PR #2437, opened 2025-06-09
ci: add `semantic.yml`

ReenigneArcher:

## Features

This PR adds a `semantic.yml` to be used with https://github.com/Ezard/semantic-prs

> [!NOTE]
> This app needs to be installed to the org and given permission to access this repository.

This can hopefully avoid the issue of contributors prefixing PR titles with spaces or other things that may be difficult to catch with a manual review.

## Notes

I use this app in https://github.com/LizardByte across every repo, although with default "types". In this case I have kept the default types as well as added the two custom ones that I know of for devicons.

This configuration will only check the PR title, and not the individual commit messages. This is the easiest approach as maintainers (or at least admins) can edit the PR title without asking the user to do it.

I didn't find any style for removing icons, let me know if there is one.

- https://github.com/devicons/devicon/blob/ca28c779441053191ff11710fe24a9e6c23690d6/.github/scripts/build_assets/util.py#L63

### DIFF
--- a/.github/semantic.yml
+++ b/.github/semantic.yml
@@ -1,0 +1,24 @@
+---
+# This is the configuration file for https://github.com/Ezard/semantic-prs
+
+enabled: true
+titleOnly: true  # We only use the PR title as we squash and merge
+commitsOnly: false
+titleAndCommits: false
+anyCommit: false
+allowMergeCommits: false
+allowRevertCommits: true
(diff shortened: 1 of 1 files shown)


Snailedlt (approved):

Snailedlt:
@canaleal Is it possible to merge this to master before the next release, or do we need to wait?

PR #2493, opened 2025-07-02
update font to use correct angular icons

weh:

## Features

Fixed version of the v2.17.0 release.

**This PR fixes the Angular issue in #2488**

## Notes

Angular Icons in the font were not the same as in the SVG files:
image

I removed `angular-plain` and `angular-plain-wordmark` from the icomoon.json and rebuilt the font files (on my local machine) using `npm run build-icons`

Now the font seems to be ok:
image


canaleal:
Looks like plain and plain-wordmark are flipped. 

The workdmark version shows plain. 

image

canaleal:
Okay, just found that only the min css needs to be updated.

canaleal:
image

canaleal (approved):

PR #2565, opened 2025-10-03
new icon: fish (original, plain)

sghng:

**This PR closes NONE**

From <https://github.com/fish-shell/fish-shell/discussions/11825#discussioncomment-14585740>


icons/fish/original.svg
```diff
@@ -0,0 +1,53 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
weh: Please check the Wiki examples for 
- [original](https://github.com/devicons/devicon/wiki/Creating-original-Icons)
- [plain](https://github.com/devicons/devicon/wiki/Creating-plain-Icons)
- [optimizing](https://github.com/devicons/devicon/wiki/Optimizing-icons)

The svg need to be on one line with a second empty line. Also it needs to be optimized using e.g. SVOMG.

The requirements for usage in the Font is a plain file, which has just one path in one color.

So in your case, we would need at least an `original.svg` and a `plain.svg`.

devicon.json
```diff
@@ -4125,6 +4125,12 @@
         "color": "#DD732A",
         "aliases": []
     },
+    {
+        "name": "fish",
+        "tags": ["shell", "programming", "language", "unix"],
+        "versions": { "svg": ["original"], "font": ["original"] },
+        "color": "#4CA9F9"
+    },
```
weh, line 4257: functional wise it does not matter for json, but the whole file is formatted using multiple lines.
So please also change this, including the non used attributes. Also as mentioned for the font version, we need the "plain" version of the icon.

```
"tags": [
  "shell",
  "programming",
  ...
]
```

weh (changes requested):
Hi @sghng, 

thank you for your contribution. I have found some things we need to change before merging the icon. Please let me know if you need more infos or explanations.

Cheers Patrik

sghng:
@weh thanks for the review! I have updated the svgs and did the optimization. I noticed that in the wiki it requires the "most prominent color" from original to be used as the color for plain svg, however, what if the brand theme is a different color? In my case, I picked blue as it seems to be the primary color on the Fish website.

sghng:
plain svg just pushed (somehow forgot it)

weh:
> plain svg just pushed (somehow forgot it)

i was just writing that the file is missing when you updated the branch.
However, i missed a last small thing. The filenames should be `fish-original.svg` and `fish-plain.svg`.
Sorry i did miss that in the initial review 😊

weh (approved):
nice work @sghng, blue is ok IMHO, can still be changed easy if needed.

PR #2609, opened 2025-11-22
fix #2608

### LINKED ISSUE #2608: [BUG]: Kotlin and Nodejs icon colors are mixed

- [x] Confirm

### Bug description

Image

Modify the 'id' attribute of 'linearGradient'

### Additional information

_No response_

Yue-plus:
Fixed 123 SVG ICONS that might have color confusion when used on the same component


canaleal:
This is really good, but it will require additional reviewers. It might also mess up the build process.

weh:
I like the idea of using the filename for ID and other references. But in the current version of the PR the icons would get converted back to multi-line with no newline at the end. Also it looks like some icons are incorrect, like angular:

image

it looks like the file is missing the closing `</radialGradient>` tag.

Another one in bootstrap:
image

i did not check all icons.

Yue-plus:
image

Oh, we can't use VScode compressed SVG function, which will modify the id attribute, causing the id attribute of each SVG file to be the same, which will reproduce the problem of #2608

Yue-plus:
> I like the idea of using the filename for ID and other references. But in the current version of the PR the icons would get converted back to multi-line with no newline at the end. Also it looks like some icons are incorrect, like angular:
> 
> image
> it looks like the file is missing the closing `</radialGradient>` tag.
> 
> Another one in bootstrap: image
> 
> i did not check all icons.

I'm pretty sure the tag not being closed is a false positive from GitHub Diff, I opened the file in VScode and there were no syntax errors, and the image displayed correctly:

image

I'm looking for a way to properly remove newlines and extra spaces from a file...

Yue-plus:
@weh Now I have compressed the contents of these 123 SVG files into one line, and I checked each file. There are no syntax errors and they can be displayed correctly.

I used this script to check and modify the file:

```powershell
# powershell
# Run in the target folder: Set-Location 'c:\code\devicon'
Get-ChildItem -Path . -Recurse -Filter *.svg | ForEach-Object {
    $path = $_.FullName
    $text = Get-Content -Raw -Path $path
    # Replace newlines and tabs with spaces, merge consecutive spaces into 1 space, and remove spaces between labels
    $text = $text -replace '\r?\n|\t', ' '
    $text = $text -replace '\s{2,}', ' '
    $text = $text -replace '>\s+<', '><'
    # Save (preserve UTF8)
    Set-Content -Path $path -Value $text -Encoding UTF8
}
```

In addition, I noticed that not all SVG files are compressed. Here I show that a total of 1758 files have been modified:

image

I can open a new PR request to submit changes if needed.

weh:
Ok, i checked the diff of the files, and it looks good to me. Did not find any missing close tags etc. Also minimizes some of the existing icons as side effect. Great work @Yue-plus.

@canaleal what did you mean by `It might also mess up the build process.` ?
I think we need to merge this into `develop` instead of `master`. 
Also we need to check for this ID/ref issue in Pull Requests too, so another thing to be aware of.

Yue-plus:
> Ok, i checked the diff of the files, and it looks good to me. Did not find any missing close tags etc. Also minimizes some of the existing icons as side effect. Great work @Yue-plus.
> 
> @canaleal what did you mean by `It might also mess up the build process.` ? I think we need to merge this into `develop` instead of `master`. Also we need to check for this ID/ref issue in Pull Requests too, so another thing to be aware of.

I submitted a new PR in the `develop` branch #2633

PR #2633, opened 2026-01-18
fix: 152 SVG ICONS that might have color confusion when used on the same component #2608

Yue-plus:
Fixed 152 SVG ICONS that might have color confusion when used on the same component

Related issues and PR:
- #2608
- #2609


Yue-plus:
We need to find a way to automatically check this issue through the workflow. There are another 29 ICONS on the `develop` branch that have color confusion problems. 😂😂😂

image

PR #2667, opened 2026-02-24
new icon: TOML (original)

macmv:

This is the SVG from the spec: https://toml.io/en/. I modified it slightly to be the correct color standalone. It's quite similar to the original in the TOML github too, although that one has different colors.


weh (changes requested):
Hi @macmv, 

thanks for your pull request. There are some things you need to update here.

The Icon files should have a single line with a second empty line in the file. This can be done when you optimize the image e.g. with svgomg. I am not sure if svgo also replaces the polygon elements with a path, or if that should be done manually.

now tho use the icon in the font we also need a second file (in your case `toml-plain.svg` which has only a single svg path with a single color. I think this should not be a big thing to do with this icon. 

If you need help or have questions, check out our wiki, or just ask, we are happy to help.

